PR #3, opened 2019-10-02
Add GitHub Actions PR checks

comigor:
Add GitHub Actions PR checks to run `pub test` and linter on all packages.


klavs:
Thanks @comigor!

I was thinking about this, but never got to actually doing it.

PR #14, opened 2019-10-04
improved context and http link

klavs:


klavs:
@comigor Would you mind reviewing the changes to `package:gql` and `package:gql_http_link`?

gql/lib/execution.dart
```diff
@@ -199,27 +199,55 @@ class Request {
...
+/// Any class extending [ContextEntry] may be added to the [Context].
+///
+/// Context entries are a subject to [Request] and [Response] equality comparisons.
+/// Only the values returned by [fieldsForEquality] will be treated as significant.
+@immutable
+abstract class ContextEntry {
+  const ContextEntry();
+
+  /// List of values to be used for equality.
+  List<Object> get fieldsForEquality;
```
comigor: `equatable` implements a similar strategy for == and hash using mixins. Maybe you can use it instead

gql_http_link/lib/gql_http_link.dart
```diff
@@ -20,94 +55,101 @@ import "package:meta/meta.dart";
...
+
+      response = parser.parseResponse(responseBody as Map<String, dynamic>);
+    } on Exception catch (e) {
+      throw ParserException(
+        originalException: e,
+        response: httpResponse,
+      );
+    }
+
+    if (httpResponse.statusCode >= 300 ||
```
comigor: Status codes >= 400 < 500 are technically client errors. Do you think a new Exception should be thrown to make it easier to handle? (Or renaming ServerException to HTTPException or something like this)

gql/lib/execution.dart
klavs: I do not exactly understand the pros of `equatable`, so I'm hesitant to use it.

Also, I'd like to remind `Link` implementers that they have to think about the `ContextEntry` equality. Exposing this getter will force implementers to think about it. Plus, this way we can transparently switch to `equatable` if needed.

gql_http_link/lib/gql_http_link.dart
klavs: This time I'm going with an equal behavior to `apollo-link-http` regarding the status codes.

I'm going with `ServerException` because I'm thinking it (and other exceptions) would eventually used across many kinds of links. For example, Link which resolves queries on the client-side could use `ServerException` to say that there was an exception in the GraphQL server (located in the client) when some exception happens during query execution.

klavs:
@comigor thank you for the review!

PR #15, opened 2019-10-16
WIP: Code Generators

klavs:
Extends the set of code builders available.

### ast_builder
Generates an AST representation of a GraphQL Document

### data_builder
Generates a typed view on top of `data` JSON. The generated classes are not supposed to be used directly as that would introduce tight coupling between your GraphQL documents and your application code.
```dart
class $ListPokemon {
  const $ListPokemon(this.data);

  final Map<String, dynamic> data;

  List<$ListPokemon$pokemons> get pokemons => (data["pokemons"] as List)
      .map((dynamic e) => $ListPokemon$pokemons(e as Map<String, dynamic>))
      .toList();
}

class $ListPokemon$pokemons {
  const $ListPokemon$pokemons(this.data);

  final Map<String, dynamic> data;

  String get id => data["id"] as String;
  String get name => data["name"] as String;
}
```

### op_builder
Wraps AST and operation name into an `Operation`.
```dart
import 'package:gql_exec/gql_exec.dart';
import 'find_pokemon.ast.gql.dart' as find_pokemon;

const FindPokemon = Operation(
  document: find_pokemon.document,
  operationName: 'FindPokemon',
);
```

### req_builder
Extend `Request` class to use specific `Operation` and provide ability to build variables. Builder pattern let's you handle nullable variables correctly (GraphQL spec pretty much requires both `undefined` and `null` values).
```dart
import 'package:gql_exec/gql_exec.dart' as _i1;
import 'find_pokemon.op.gql.dart' as _i2;

class FindPokemon extends _i1.Request {
  FindPokemon()
      : super(operation: _i2.FindPokemon, variables: <String, dynamic>{});

  set name(String value) => variables["name"] = value;
}
```

### enum_builder
Generates an enum-like class per GraphQL enum type. Defines known enum values to be used in your code, and allows unknown enum values to be used without causing runtime error when handling response data.
```dart
class ReleaseType {
  const ReleaseType(this.value);

  final String value;

  static const ReleaseType ALPHA = ReleaseType('ALPHA');

  static const ReleaseType BETA = ReleaseType('BETA');

  static const ReleaseType GAMMA = ReleaseType('GAMMA');

  @override
  int get hashCode => value.hashCode;
  @override
  bool operator ==(Object o) => o is ReleaseType && o.value == value;
}
```

### scalar_builder
Generates a container for a scalar value to be used when viewing the response data and building request variables.
```dart
class ID {
  const ID(this.value);

  final String value;

  @override
  int get hashCode => value.hashCode;
  @override
  bool operator ==(Object o) => o is ID && o.value == value;
}
```

### input_builder
Generates an input builder to be used to build request variables.
```dart
class MutationInput {
  final Map<String, dynamic> input = <String, dynamic>{};

  set a(String value) => input['a'] = value;
}
```

### schema_builder
Combines `enum_builder`, `input_builder` and `scalar_builder`.

## Usage example
```dart
final result = await link.request(
  FindPokemon()..name = "Charizard",
);
final data = $FindPokemon(result.data);

final pokemon = data.pokemon;

print("Found ${pokemon.name}");
print("ID: ${pokemon.id}");

final weight = pokemon.weight;
final height = pokemon.height;

print(
  "Weight: ${weight.minimum} – ${weight.maximum}",
);
print(
  "Height: ${height.minimum} – ${height.maximum}",
);
````

### CODE REVIEW

**Comment by klavs** [comment]
I'm at a point where I'm happy with the set of builders I have and the core concepts the generated classes use.

Now I just have grind to get it release-ready.
The things that are still missing:
- Fragments in Data Viewer
- All field types in Data Viewer
- Types from schema in Request Builder

@micimize, @comigor, both of you are working with code generation. Would you be able to review the description of this PR to maybe point out the things I have not thought about?

**Comment by micimize** [comment]
This is exciting stuff! I think the main thing to me is that the more extensible/discretely customizable it is the better. Like, with the  `Visitor` pattern, a user can take a preexisting visitor, override one method, and still take advantage of all the preexisting work quite easily.

One issue with the wrapped map approach as in `data_builder` is that when you implement non-primitive types, it looks like they'll be cast every read, which could be undesirable for `DateTime`, custom scalars, etc.

Also, kinda a nit-pick but I think it'd be good to [maintain the editor tooling around enums](https://gist.github.com/micimize/39f6d59492b3c6792e607a476fb131a6) somehow or another.

FWIW: In `graphql-to-dart`, I ended up trying to model the actual graphql "inheritance chain" pretty closely. The main trick was to have a [protected `fields` attribute](https://github.com/gql-dart/gql/blob/23fd0a4df74a106d9be53a752dc0c37d9925947e/gql_code_gen/lib/schema_builder.dart) on schema types that operation data classes expose. Time will tell if doing so was worthwhile 😅

**Comment by comigor** [comment]
Wow, that's nice!

Why there's dollar signs on `data_builder`? Just to "warn" users not to use it directly?
Why don't you use some lib like equatable to avoid the boilerplate of overriding `hashCode` and `operator ==`?

**Comment by klavs** [comment]
@micimize 
> I think the main thing to me is that the more extensible/discretely customizable it is the better.
I'm actually thinking of splitting the builder (used by `build_runner`) off to a separate package from the code builder / code gen so that it's possible to use the code gen in any custom combination. 

> Like, with the `Visitor` pattern, a user can take a preexisting visitor, override one method, and still take advantage of all the preexisting work quite easily.
The interesting thing I found about the `Visitor` pattern is that sometimes you want to visit same node multiple times. Say, once to add a field to a class and once to add that same field to the constructor. So I ended up moving away from that pattern.

> One issue with the wrapped map approach as in `data_builder` is that when you implement non-primitive types, it looks like they'll be cast every read, which could be undesirable for `DateTime`, custom scalars, etc.
That's a great point! I think I could work around that by caching the casted value when the getter is called.

> Also, kinda a nit-pick but I think it'd be good to maintain the editor tooling around enums somehow or another.
This is a good suggestion. I think I'm not yet there to comprehend the effect of such pattern on the usage of it. I have to keep this in mind.

> FWIW: In graphql-to-dart, I ended up trying to model the actual graphql "inheritance chain" pretty closely. The main trick was to have a protected fields attribute on schema types that operation data classes expose. Time will tell if doing so was worthwhile 😅
I have my doubts about that approach but I hope you succeed and make my doubts go away.

@comigor 
> Why there's dollar signs on data_builder? Just to "warn" users not to use it directly?
Exactly. I'm hoping the type inference let's users avoid using those types directly, but when they need to use them, it clearly look odd suggesting it's not good to do so. In my experience, it's really bad to use GraphQL types in component/modules not related to data access, because it makes them fragile.

> Why don't you use some lib like equatable to avoid the boilerplate of overriding `hashCode` and `operator ==`?
I'm trying to avoid adding extra dependencies in both the generated code and the packaged code.



If you guys want to help me get this done, I can prepare the code for it and come up with concrete work items and maybe some test cases.
It might take another 3 months for me to finish it.
As I mentioned in one of the Artemis issues, I'm open to you steeling/using the end result under your own projects. Whatever makes the ecosystem better is good to me.

**Comment by klavs** [comment]
This is now published as `gql_code_gen: v0.2.0-alpha.1` (no support for fragments).
Feel free to try it out and file issues about unexpected behavior.

**Comment by klavs** [comment]
Usage example can be found in this branch. See `gql_example_cli`

**Comment by smkhalsa** [comment]
For fragments, may I suggest that we generate them as a concrete class and then have any queries that use the fragment simply implement that class? That way the fragments class can be used independent of any particular query.

For example, I have a music player app, and I want my player to interact with `Song` objects that may be fetched using a number of different queries. The above approach would allow the player to accept any `Song` subclass.

**Comment by smkhalsa** [comment]
@klavs I'm just created a [fragment builder](https://github.com/smkhalsa/gql/commit/e091789a9eaf40ab2fec2ed08740a109018e0eec) that creates a single `fragment.gql.dart` file with fragment classes for all fragments defined in any `.graphql` within the lib folder.

Let me know what you think. Next steps would be to:
1. update the data builder to implement any necessary fragment classes
2. include fields from fragment class in data class

**Comment by smkhalsa** [comment]
> 1. update the data builder to have data classes implement any necessary fragment class interfaces

[implemented](https://github.com/smkhalsa/gql/commit/39cc22c1a53df4dfed858dbb663fe7ddd1486b51)

**Comment by smkhalsa** [comment]
> - [x] [update the data builder to have data classes implement any necessary fragment class interfaces](https://github.com/smkhalsa/gql/commit/39cc22c1a53df4dfed858dbb663fe7ddd1486b51)

I just realized that my implementation above only works for the top level data class. However, consider a situation like this:

```graphql
fragment pokemonWeights on Pokemon {
  weight {
    id
    minimum
  }
}

query FindPokemon($name: String!) {
  pokemon(name: $name) {
    ...pokemonWeights
    weight {
      minimum
      maximum
    }
    height {
      minimum
      maximum
    }
  }
}
```

In this case, not only should `$FindPokemon$pokemon` implement the `$pokemonWeights` interface, `$FindPokemon$pokemon$weight` should also implement the `$pokemonWeights$weight` interface.

I'll dig into this a bit deeper.

klavs:
To me, practice of reusing fragments promotes and invites overfetching. But I might think about it that way because I use a certain flavour of apollo-inspired practices. It could be that there are other valid ways to think about it.

@smkhalsa, could you adjust the `gql_example_cli` to include an example of how your idea would look like from the users perspective?

smkhalsa:
@klavs, I've just added a [flutter example](https://github.com/smkhalsa/gql/tree/fragment-builder/gql_example_flutter) to show how fragments could be used.

To me, fragments are all about code reuse, so I felt an example using flutter would be more instructive. As you can see, both the [`AllPokemonScreen`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/all_pokemon/all_pokemon.dart) and the [`PokemonDetailScreen`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_detail/pokemon_detail.dart) use the [`PokemonCard`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_card/pokemon_card.dart) widget.

Using a fragment, we can define all of the fields that [`PokemonCard`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_card/pokemon_card.dart) needs in the [`PokemonCard`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_card/pokemon_card_fragment.graphql) fragment then include it in the queries from [`AllPokemonScreen`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/all_pokemon/all_pokemon.dart) and [`PokemonDetailScreen`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_detail/pokemon_detail.dart). This has several significant advantages:

1. We colocate the data needed for [`PokemonCard`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_card/pokemon_card.dart) next the the widget that uses it.
2. If the data needed for [`PokemonCard`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_card/pokemon_card.dart) changes, we only need to update it in the [`PokemonCard`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_card/pokemon_card_fragment.graphql) fragment, not in the queries that use the fragment. (NOTE: since we haven't yet updated `data_builder` to automatically include getters for implemented fragments, I've included all the fragment fields in the queries.)
3. Since we are building the fragment as a Class, our [`PokemonCard`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_card/pokemon_card_fragment.graphql) can simply [accept](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_card/pokemon_card.dart#L6) a [`$PokemonCard`](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/fragments.gql.dart) class, allowing any parent widget to pass in data from any query that implements the fragment.

I hope this is clear. Please let me know your thoughts.

klavs:
@smkhalsa looks good and I would like to encourage you to continue exploring this.

Besides some smaller things, I'm wondering would do nested and multiple levels deep fragments behave.

klavs:
I think the phase one could be to generate fragment classes separately and use them by passing data via the constructor instead of data classes implementing fragment classes.

The reason for it is that from the standpoint of data classes, we have to consider anonymous fragments as well. Also anonymous classes are usually used with unions and interfaces, which are still not considered by the data classes. I'm still not sure how they impact the API. I'm hoping they could guide us in understanding what the API should be.

smkhalsa:
> I'm wondering would do nested and multiple levels deep fragments behave

@klavs great timing :) I just [pushed an update](https://github.com/smkhalsa/gql/commit/64544c1a42c5ad211ae75d471fb0b9172f13a1ca) that now adds fragment getters to data classes and also adds [recursive fragment support](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_example_flutter/lib/src/pokemon_card/pokemon_card_fragment.graphql).

smkhalsa:
> from the standpoint of data classes, we have to consider anonymous fragments as well.

For inline fragments, since we don't know at build time whether the class is of the given type, I just [add all inlined fields](https://github.com/smkhalsa/gql/blob/fragment-builder/gql_code_builder/lib/src/operation/data.dart#L137) to the data class. (I still need to test this)

I thought about creating a subclass for each concrete type used in the inline fragment, but I didn't really see the value in doing so (and it would add complexity).

smkhalsa:
@klavs If you have time, I'd love to for you to test out any fragment corner cases.

smkhalsa:
In order to take full advantage of the request builder and not have to manually instantiate the data class with the returned data, I'd like to recommend we add a parse function to the request class, similar to artemis.

I've taken a first pass at this [here](https://github.com/smkhalsa/gql/tree/built-request). (see [example](https://github.com/smkhalsa/gql/blob/built-request/gql_example_flutter/lib/src/all_pokemon/all_pokemon.req.gql.dart#L12))

This is necessary for me to be able to incorporate the builders into the client I'm working on.

There may be a better way to accomplish this. Please let me know your thoughts.

klavs:
@smkhalsa you can easily create a builder which gives you that kind of API.

Currently:
```dart
    final result = await link
        .request(
          FindPokemon()..name = find,
        )
        .first;

    final data = $FindPokemon(result.data);

    final pokemon = data.pokemon;
```

Proposed API:
```dart
    final result = await FindPokemon(
        (b) => b..name = find,
    ).execute(link).first;

    final pokemon = result.data.pokemon;
```

Now I'm starting to think that instead of or in addition to a `req_builder`, I create a `var_builder` so that it is actually independent of the link or client user is using it with.

klavs:
I've published these builders as 0.0.X version packages as part some of them are already useful.

@smkhalsa please create PRs with your work on fragments and built request to discuss them separately.

PR #21, opened 2019-11-14
Adopt Normalize

klavs:
- adopt Normalize https://github.com/smkhalsa/normalize
- use `gql_pedantic`
- streamline GitHub actions


klavs:
@smkhalsa, please review

PR #28, opened 2019-12-17
Recursive #import inlining

### LINKED ISSUE #27: #imports support ast_builder
Related to https://github.com/comigor/artemis/issues/49 -

I'd like to add `ast_builder` support to `graphql-to-dart`, and having it handle imports internally would simplify it considerably. Examples of this pattern in the js world: [`graphql-import`](https://github.com/Urigo/graphql-import), [`graphql-import-loader`](https://github.com/prisma-labs/graphql-import-loader)

micimize:
closes #27
 
This implements a simple solution to #27 in a recursive fashion. It probably isn't the most elegant solution to the problem, but (I'm fairly sure) it handles nested imports without risk of an infinite loop due to a cycle.

### DIFF
--- a/gql_code_gen/lib/ast_builder.dart
+++ b/gql_code_gen/lib/ast_builder.dart
@@ -34,6 +34,41 @@ Set<String> allRelativeImports(String doc) {
   return imports;
 }
 
+Future<String> inlineImportsRecursively(BuildStep buildStep) async {
+  final Map<String, String> importMap = {};
+  final Set<String> seenImports = {};
+
+  void collectContentRecursivelyFrom(AssetId id) async {
+    importMap[id.path] = await buildStep.readAsString(id);
+    final segments = id.pathSegments
+      //..removeAt(0) // strip lib
+      ..removeLast();
+
+    final imports = allRelativeImports(importMap[id.path])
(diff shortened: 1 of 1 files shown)


gql_code_gen/lib/ast_builder.dart
```diff
@@ -10,8 +13,62 @@ import "package:gql/language.dart";
...
+        .toSet();
+    for (final assetId in assetIds) {
+      await collectContentRecursivelyFrom(assetId);
+    }
+  }
+
+  await collectContentRecursivelyFrom(buildStep.inputId);
+
+  seenImports.where((i) => !importMap.containsKey(i)).forEach(
+      (missing) => log.warning("Could not import missing file $missing."));
```
klavs: please use commas liberally to make the lines collapse in more readable way 
```
foo.where(bar,).forEach(baz,);
```

micimize:
@klavs alright - I've applied the style changes and fixed an issue with the regex not capturing multiple lines. I'm not sure why the `gql_code_gen` workflow is failing, its tests pass locally.

klavs:
Merged, fixed and deployed as v0.1.5

PR #37, opened 2020-01-25
Ensure doubles are generated properly

smkhalsa:
This PR updates the data_builder to call `toDouble` on fields that expect a `double`. This ensures that even if the data received is stored in the `data` map as an integer, it will still get returned as a `double`.

### DIFF
--- a/gql_code_builder/lib/src/operation/data.dart
+++ b/gql_code_builder/lib/src/operation/data.dart
@@ -223,7 +223,9 @@ Method _buildGetter(
                 .code
             : node.selectionSet == null
                 ? fieldTypeDef == null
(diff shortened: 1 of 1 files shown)


gql_code_builder/lib/src/operation/data.dart
```diff
@@ -223,7 +223,9 @@ Method _buildGetter(
                 .code
             : node.selectionSet == null
                 ? fieldTypeDef == null
-                    ? dataField.asA(returns).code
+                    ? returns.symbol == "double"
```
klavs: Ideally, there should be a generalized coercer functionality that let's this problem be solved for custom types as well.

But for now this is good enough.

Issue #40, opened 2020-01-27
[gql_build] Build data for Mutations


smkhalsa:
As you can see [here](https://github.com/gql-dart/gql/blob/master/gql_code_builder/lib/src/operation/data.dart#L35), we are currently only building data classes for `Query`s, not `Mutation`s.

klavs:
True. It should be quite simple to fix by mapping operation type to correct object name. The only thing to remember is that `{ fooBar }` is a valid query and then the operation type (I think) is `null`.

We can ignore for now the fact that there might be a custom schema definition:
```
schema {
  query: OtherQuery
  mutation: SomeMutation
  subscription: SuperScription
}
```

smkhalsa:
> `{ fooBar }` is a valid query and then the operation type (I think) is `null`

Actually, shouldn't the operation type in that case be `query`?

smkhalsa:
closed in #41

PR #41, opened 2020-01-27
Allow mutation and subscription operations

smkhalsa:
Per https://github.com/gql-dart/gql/issues/40, this PR maps the operation to the appropriate type.

### DIFF
--- a/gql_code_builder/lib/src/operation/data.dart
+++ b/gql_code_builder/lib/src/operation/data.dart
@@ -42,9 +42,9 @@ String _operationType(
   DocumentNode schema,
   OperationDefinitionNode op,
 ) {
-  final schemaDef = schema.definitions.whereType<SchemaDefinitionNode>().first;
-  if (schemaDef == null) return defaultRootTypes[op.type];
-  return schemaDef.operationTypes
+  final schemaDefs = schema.definitions.whereType<SchemaDefinitionNode>();
+  if (schemaDefs.isEmpty) return defaultRootTypes[op.type];
+  return schemaDefs.first.operationTypes
       .firstWhere((opType) => opType.operation == op.type)
       .type
       .name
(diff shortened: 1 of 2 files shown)


smkhalsa:
@klavs and this as well?

gql_code_builder/lib/src/operation/data.dart
```diff
@@ -32,9 +32,23 @@ List<Class> _buildOperationDataClasses(
...
+        op,
+      ),
     );
 
+String _operationType(
+  DocumentNode schema,
+  OperationDefinitionNode op, [
+  Map<OperationType, String> rootTypes = defaultRootTypes,
+]) =>
+    schema.definitions
```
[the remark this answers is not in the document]
smkhalsa: 😳you're right.

I've updated it and added support for a custom schema definition.

gql_code_builder/lib/src/operation/data.dart
```diff
@@ -32,9 +32,25 @@ List<Class> _buildOperationDataClasses(
...
+        schema,
+        op,
+      ),
     );
 
+String _operationType(
+  DocumentNode schema,
+  OperationDefinitionNode op,
+) {
+  final schemaDefs = schema.definitions.whereType<SchemaDefinitionNode>();
```
klavs: I wish Dart analyzer had a rule for empty line before return.

```suggestion:-0+7
  final schemaDefs = schema.definitions.whereType<SchemaDefinitionNode>();
  
  if (schemaDefs.isEmpty) return defaultRootTypes[op.type];
  
  return schemaDefs.first.operationTypes
      .firstWhere(
        (opType) => opType.operation == op.type,
      ).type.name.value;
```

smkhalsa: done

PR #55, opened 2020-02-16
Merge fragment builder into data builder

smkhalsa:


gql_code_builder/lib/src/operation/data.dart
```diff
@@ -3,48 +3,85 @@ import "package:built_collection/built_collection.dart";
 import "package:code_builder/code_builder.dart";
 import "package:gql/ast.dart";
 import "package:gql_code_builder/src/common.dart";
+import "package:gql_code_builder/source.dart";
+
+class SourceSelections {
```
smkhalsa: I needed to keep track of the source URL along with the selections. If you feel like there's a better way to do this, please let me know.

gql_code_builder/lib/src/operation/data.dart
```diff
@@ -115,16 +139,17 @@ List<Class> _buildSelectionSetDataClasses({
         ..implements = ListBuilder(
             superclassSelections.keys.map<Reference>((superName) => refer(
                   superName,
-                  superName == parentName ? null : fragmentsDocUrl,
+                  superclassSelections[superName]
+                      .url
+                      ?.replaceAll(RegExp(r".graphql$"), ".data.gql.dart"),
```
smkhalsa: I temporarily hard-coded the string replacement here. Please let me know how you think we should pass in the extension.

gql_code_builder/lib/src/operation/data.dart
```diff
@@ -358,15 +378,18 @@ Method _buildGetter(
...
     );
     final typeMap = {
       ...defaultTypeMap,
       if (node.selectionSet != null)
         typeName: refer("$prefix\$$name")
       else if (fieldTypeDef != null)
-        typeName: refer(typeName, schemaUrl)
+        typeName: refer(
+          typeName,
+          schemaSource.url.replaceAll(RegExp(r".graphql$"), ".schema.gql.dart"),
```
smkhalsa: Same thing here - how do you suggest we pass the extension?

smkhalsa:
@klavs I've successfully merged fragment builder into data builder, and it is working correctly in both the flutter example and my own project.

gql_code_builder/lib/src/operation/data.dart
klavs: I think this makes sense. Please make it private.

gql_code_builder/lib/src/operation/data.dart
klavs: I'm trying to think of a way to solve this issue by using the custom import Allocator.

Please add a code comment here about it being hard-coded. I'll try to fix this when improving the allocator.

gql_code_builder/lib/src/operation/data.dart
klavs: Just add a code comment for now.

Issue #5, opened 2019-10-03
Implement UniqueTypeNames validation rule


klavs:
Implement UniqueTypeNames validation rule ([ref](https://github.com/graphql/graphql-js/blob/master/src/validation/rules/UniqueTypeNames.js))

Unlike `graphql-js`, there is no validation context given when validating a document. A schema validation rule should be self-sufficient and based only on the given AST document.
See https://github.com/gql-dart/gql/tree/master/gql/lib/src/validation.

jonas-jonas:
I opened #56. If you have any feedback, I'm happy to incorporate that into the PR.

PR #56, opened 2020-02-16
Implement UniqueTypeNames rule

jonas-jonas:
I saw your post on Reddit about these validation rules, and looked around the currently implemented 
 validations and the reference implementation and came up with this. Hope this helps.

Closes https://github.com/gql-dart/gql/issues/5.

### DIFF
--- a/gql/lib/src/validation/validator.dart
+++ b/gql/lib/src/validation/validator.dart
@@ -107,7 +107,7 @@ ValidatingVisitor _mapRule(ValidationRule rule) {
     case ValidationRule.uniqueOperationTypes:
       return UniqueOperationTypes();
     case ValidationRule.uniqueTypeNames:
-    return UniqueTypeNames();
+      return UniqueTypeNames();
     default:
       return null;
   }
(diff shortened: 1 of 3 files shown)


gql/lib/src/validation/validator.dart
```diff
@@ -103,6 +106,8 @@ ValidatingVisitor _mapRule(ValidationRule rule) {
       return LoneSchemaDefinition();
     case ValidationRule.uniqueOperationTypes:
       return UniqueOperationTypes();
+    case ValidationRule.uniqueTypeNames:
+    return UniqueTypeNames();
```
klavs: A tiny formatting issue here.

jonas-jonas: Thanks, fixed.

jonas-jonas:
> At this point I care about completeness, so this is great even if I my doubts about the design I've set up here.

I guess the current system works reasonably well, for now. Anyway, I resolved the whitespace issue.

What about the other validators? I might have some more time on hand the next days. Should I just open another PR or comment on the ones, I'm interested in?

klavs:
@jonas-jonas, thank you very much!

Comment on an issue first (just in case someone might want to try their hand as well) and implement without any hesitation just like you did with this one.

PR #58, opened 2020-02-17
Add UniqueInputFieldNames validator

jonas-jonas:
This adds the UniqueInputFieldNames validator. It required a few more tweaks to the original JS code, because as far as I can tell, there is no "enter" and "leave" hooks for the visitors (yet?). 

I also copied most of the test cases from the JS source to make sure that this implementation is complying with the reference.


klavs:
@jonas-jonas, thank you!

Your observations about our visitor implementation is correct, so it's ok to tweek the implementation.

Issue #62, opened 2020-02-22
[gql_build] regression for interfaces


smkhalsa:
https://github.com/gql-dart/gql/commit/2ec34002ed4771c628462f432fb1bb03c66eba17 introduced a regression for queries of GraphQL interfaces.

@klavs, I haven't reviewed the changes in much detail. Any ideas where this may have been introduced?

smkhalsa:
Here's the error that's thrown:

```
NoSuchMethodError: The method '+' was called on null.
Receiver: null
Tried calling: +("#data")
[SEVERE] gql_build:data_builder on lib/services/graphql/audio_history_query.graphql (cached):
```

smkhalsa:
We should probably add some unit tests to avoid these types of regressions in the future. When I first started researching dart builders, I looked a bit at build_test, but I haven't dug deep into it yet.

If you have any sample tests for the builders, it'd be helpful to review them.

klavs:
Please add `--verbose` flag to see the line failing.

smkhalsa:
Here's the verbose output:

```
[SEVERE] gql_build:data_builder on lib/services/graphql/audio_history_query.graphql (cached):

NoSuchMethodError: The method '+' was called on null.
Receiver: null
Tried calling: +("#data")
dart:core                                                               Object.noSuchMethod
package:gql_code_builder/src/operation/data.dart 143:51                 _buildSelectionSetDataClasses.<fn>.<fn>
dart:core                                                               new List.from
package:built_collection/src/list/list_builder.dart 52:19               ListBuilder.replace
package:built_collection/src/list/list_builder.dart 28:45               new ListBuilder
package:gql_code_builder/src/operation/data.dart 139:24                 _buildSelectionSetDataClasses.<fn>
package:code_builder/src/specs/class.g.dart 304:33                      _$ClassBuilder.update
package:code_builder/src/specs/class.g.dart 44:28                       new _$Class
package:gql_code_builder/src/operation/data.dart 136:5                  _buildSelectionSetDataClasses
package:gql_code_builder/src/operation/data.dart 186:31                 _buildSelectionSetDataClasses.<fn>
dart:_internal                                                          ExpandIterator.moveNext
package:gql_code_builder/src/operation/data.dart 185:51                 _buildSelectionSetDataClasses
package:gql_code_builder/src/operation/data.dart 165:22                 _buildSelectionSetDataClasses.<fn>
dart:_internal                                                          ExpandIterator.moveNext
[...]

PR #64, opened 2020-02-22
take care of nullable url

klavs:
closes https://github.com/gql-dart/gql/issues/62


klavs:
already published to pub

Issue #65, opened 2020-02-22
Error when parsing SDL schema


comigor:
Hey @klavs! I'm getting an error while using the language SDL parser (`parseString`) on some of my examples. It say it can't find the Entity interface while parsing the [Area object](https://gist.github.com/comigor/cc4664e68499e7f330064a5543736371#file-graphbrainz-schema-graphql-L32) even though [it's declared](https://gist.github.com/comigor/cc4664e68499e7f330064a5543736371#file-graphbrainz-schema-graphql-L1114) on the same schema:
```
Error on line 32, column 28 of lib/graphbrainz.schema.graphql: Unknown definition type 'Entity'
   ╷
32 │ type Area implements Node, Entity {
   │                            ^^^^^^
   ╵
package:gql/src/language/parser.dart 175:5   _Parser._parseDefinition
package:gql/src/language/parser.dart 117:22  _Parser._parseMany
package:gql/src/language/parser.dart 142:22  _Parser._parseDocument
package:gql/src/language/parser.dart 41:27   _Parser.parse
package:gql/src/language/parser.dart 17:17   parse
package:gql/src/language/parser.dart 27:5    parseString
```

The full schema, for referece: [graphbrainz.schema.graphql](https://gist.github.com/comigor/cc4664e68499e7f330064a5543736371).

comigor:
And another problem when parsing the [GitHub schema](https://gist.github.com/comigor/5be2f6172e70aa118d99f86f272022a8):
```
Bad state: No element
```

klavs:
@comigor, it should be `type Area implements Node & Entity`

klavs:
@comigor, I was able to parse the official github schema just fine. Would you be able to narrow your SDL to the part which actually fails?

klavs:
It's line 813.
https://gist.github.com/comigor/5be2f6172e70aa118d99f86f272022a8#file-github-schema-graphql-L810-L814

comigor:
Oh, you're absolutely right about the spec, though some tools that convert Introspection queries to SDL seem to generate the SDL with commas.

Do you know if the spec was different before 2018? If it used a comma before, shouldn't we be retro-compatible and parse it as well?

Reference of code used in the conversion:  https://gist.github.com/stubailo/041999ba5b8b15cede60b93ff9a38f53#file-introspection-result-to-sdl-js

comigor:
[...]

PR #66, opened 2020-02-22
Peek for opening bracket while parsing interfaces

comigor:
I'm not sure that's the ideal code, but worked with [my example](https://github.com/gql-dart/gql/issues/65).

Solves the first stacktrace of https://github.com/gql-dart/gql/issues/65


klavs:
@comigor, please add a test to how what exact problem you are solving

comigor:
Let's discuss what we want to do on the issue I've just opened, and if it's the case, I'll reopen this PR.

PR #74, opened 2020-02-23
fragment generation tests

klavs:
- add `__typename` field to object and interface definitions
- add some fragment-related tests


gql_example_build/test/fragments/shape_test.dart
```diff
@@ -0,0 +1,100 @@
...
+    return "$type(area: $area, sideLengthA: ${shape.sideLengthA}, sideLengthB: ${shape.sideLengthB})";
+  }
+}
+
+String getShapeInfoNoRefs($Shape data) {
+  final shape = data.shape;
+  final type = shape.$__typename;
+  final area = shape.area;
+
+  final square = shape.asSquare();
```
klavs: @smkhalsa, here's the API I originally imagined. What do you think? I like this approach because it lets user avoid using generated classes directly.

smkhalsa: One minor note: if we did this, we'd probably want to prefix the `asSubclass()` methods with `$` to avoid collisions with data fields.

What I like about using the generated subclasses directly is that `if (class is classAsSubclass)` feels intuitive and doesn't require any additional typecasting.

That said, if we did go this route, it would feel more intuitive to me to also have a `$isSubclass` property rather than checking for null with the `$asSubclass()` method.

```dart
if (shape.$isSquare) {
  final square = shape.$asSquare();
  return "$type(area: $area, sideLength: ${square.sideLength})";
}

if (shape.$isRectangle) {
  final rectangle = shape.$asRectangle();
  return "$type(area: $area, sideLengthA: ${rectangle.sideLengthA}, sideLengthB: ${rectangle.sideLengthB})";
}
```

PR #75, opened 2020-02-24
Make GraphQLError implement Exception

smkhalsa:
This should make it easier for clients (like Ferry) to provide a `List<Exception>` which could include both Network Errors and GraphQL Errors.


smkhalsa:
After further consideration, I'm closing this as I feel it is conflating multiple concepts.

PR #77, opened 2020-02-25
Remove duplicate ast imports

### LINKED ISSUE #73: [gql_build] duplicate ast imports
I'm running into an issue with the new import functionality. If I have a query that uses multiple fragments which themselves use a common fragment, the `ast_builder` will include multiple instances of the import for that common fragment in its List of definitions.

@klavs I assume we can simply check for uniqueness [here](https://github.com/gql-dart/gql/blob/master/gql_code_builder/lib/ast.dart#L22), but let me know your thoughts.

smkhalsa:
Fixes #73.

Looks like it was a very simple fix.

### DIFF
--- a/gql_build/CHANGELOG.md
+++ b/gql_build/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.0.9
+
+- add `__typename` field to interface and object types
+
 ## 0.0.8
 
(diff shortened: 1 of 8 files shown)


smkhalsa:
@klavs done

PR #84, opened 2020-04-21
Add a GENERATED CODE header

liyuqian:
So it's clear that such code is auto-generated and should not be
modified by hand.


klavs:
Thank you @liyuqian!

PR #86, opened 2020-04-23
gql/schema.dart and gql/operation.dart

### LINKED ISSUE #25: Typed graphql document visitor
It would make some use cases easier to implement if there was a type-aware visitor provided. Maybe something that visits AST nodes that have a `type` attribute?

I'm basically trying to implement something like what I talk about in https://github.com/comigor/artemis/issues/39, where a base `schema_types.dart` file gets generated with just the subset of schema types referenced by the collected operations and fragments.

micimize:
Closes #25

As discussed in #25, this extracts the `schema` and `operation` definition modeling work I did in [`major_graphql_generator`](https://github.com/micimize/major/tree/master/major_graphql_generator) into `gql` subpackages.

They are currently very much geared towards my code gen use-case:
* `schema` does not have any field resolution like [`graphql_server`](https://pub.dev/packages/graphql_server) does, or much validation
* They are mostly modeled as wrappers around the top-level ast nodes.

However I _think_ they're useful enough in their own right, as well as a decent starting point for more fully featured libraries.

TODOS before `Ready for Review`::
- [X] rigorous `gql/schema.dart` docs
- [X] rigorous `gql/operation.dart` docs
- [X] snapshot-like tests


micimize:
@klavs this is ready for review!

I added `equatable` as a dependency to make it so two definitions based on the same ast will be equivalent, as they are lazily resolved which could lead to confusing errors.

gql/lib/src/schema/schema.dart
```diff
@@ -0,0 +1,200 @@
...
+    ObjectTypeDefinition objectType,
+  ) =>
+      AbstractType.isPossibleType(abstractType, objectType);
+
+  /// Build a schema from [documentNode]
+  static GraphQLSchema fromNode(DocumentNode documentNode) =>
+      buildSchema(documentNode);
+}
+
+/// Build a [GraphQLSchema] from [documentNode]
```
klavs, line 115:
```suggestion
/// Build a [GraphQLSchema] from [documentNode].
```

gql/lib/src/schema/schema.dart
```diff
@@ -0,0 +1,200 @@
...
+/// The resulting schema has no `resolve` methods
```
klavs, line 116: Could you give state the reason for it in the docs?

gql/lib/src/schema/definitions/base_types.dart
```diff
@@ -0,0 +1,264 @@
...
+/// * [NamedType]
+/// * [ListType]
+/// * [Argument]
+/// * [Directive]
+/// * [TypeSystemDefinition]
+/// * [TypeDefinition]
+part of "definitions.dart";
+
+@immutable
+abstract class GraphQLEntity extends Equatable {
```
klavs, line 14: No need to extends `Equatable`.

You can use `DeepCollectionEquality` to implement `operator==` and `hashCode`. See `Node` in `lib/src/ast/ast.dart`.

After that, the dependency on `package:equatable` will be redundant.

klavs:
Oh... After making the comment about scalars I noticed that it's taken from the spec.

It's weird they chose that wording. Feel free to stick to it.

klavs:
Regarding `package:equatable`. Their implementation appears to be faster than `DeepCollectionEquality`. But I would still prefer not adding another dependency for the same job before doing a bigger gql-focused investigation.
```
----Comparing Identical----
RAW:
Template(RunTime): 0.07521036055809247 us.
DeepCollectionEquality:
Template(RunTime): 0.07585805319712868 us.
Equatable:
Template(RunTime): 0.08521141057477866 us.
Empty Equatable:
Template(RunTime): 0.08343149744269945 us.
----Comparing Equal----
RAW:
Template(RunTime): 0.17187927818265852 us.
DeepCollectionEquality:
Template(RunTime): 0.7587708216198088 us.
Equatable:
Template(RunTime): 0.3789735614779438 us.
Empty Equatable:
Template(RunTime): 0.15959128034737916 us.
----Comparing Different----
RAW:
Template(RunTime): 0.20565169943880737 us.
DeepCollectionEquality:
Template(RunTime): 0.8897520308204248 us.
Equatable:
Template(RunTime): 0.4655652788864051 us.
----Instantiation----
RAW:
Template(RunTime): 0.1025187472188586 us.
DeepCollectionEquality:
Template(RunTime): 0.11835629149497608 us.
Equatable:
Template(RunTime): 0.09798510773340113 us.
Empty Equatable:
Template(RunTime): 0.09261688737425289 us.
----Hash code----
RAW:
Template(RunTime): 0.150860997657355 us.
DeepCollectionEquality:
Template(RunTime): 1.1773993485449405 us.
Equatable:
Template(RunTime): 0.5863965958504818 us.
Empty Equatable:
Template(RunTime): 0.21492783421842407 us.

```

klavs:
I used `DeepCollectionEquality` inefficiently.
Here's a more efficient way (very close to what equatable actually does). I'll be fixing it in an upcoming PR.

```
  @override
  bool operator ==(Object o) {
    if (identical(this, o)) return true;

    if (o.runtimeType == runtimeType) {
      return const ListEquality<Object>(
        DeepCollectionEquality(),
      ).equals(
        (o as Node)._children,
        _children,
      );
    }

    return false;
  }

  @override
  int get hashCode => const ListEquality<Object>(
        DeepCollectionEquality(),
      ).hash(
        _children,
      );
```

gql/README.md
```diff
@@ -189,6 +191,76 @@ void main() {
...
 
+## `package:gql/schema.dart` (experimental)
+
+Higher-level schema type definitions derived from the `gql/ast.dart` asts.
+Adds convenience attributes and type dereferencing facilities.
+
+**NOTE**: does not currently have field resolution.
+
+```dart
+import "package:gql/language.dart" as lang;
```
klavs, line 202: Please put these in `example/` as well. There are some `dartfmt` and `dartanalyzer` issues here.

gql/lib/src/schema/definitions/definitions.dart
```diff
@@ -0,0 +1,403 @@
...
+*/
+
+/// A [field](https://spec.graphql.org/June2018/#sec-Language.Fields)
+/// describes one discrete piece of information available to request within a selection set.
+///
+/// Some fields describe complex data or relationships to other data.
+/// In order to further explore this data, a field may itself contain a selection set,
+/// allowing for deeply nested requests.
+///
+/// All GraphQL operations must specify their selections down to fields which return scalar values
```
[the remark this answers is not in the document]
micimize: This is also copied from the spec - idk why but they very often refer to `{ enums, lists, scalars }` this way

gql/lib/src/schema/definitions/definitions.dart
```diff
@@ -0,0 +1,403 @@
...
+        return "mutation";
+      case OperationType.subscription:
+        return "subscription";
+    }
+    throw StateError("Impossible OperationType $this");
+  }
+}
+
+/*
+  Begin [TypeResolver]-enabled classes
```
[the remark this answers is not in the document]
micimize: maybe so. I'm not a big fan of huge files like this, but there isn't much good middle ground here between "all together" or "almost 1 definition per-file". Would rather refactor to the latter... eventually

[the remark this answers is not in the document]
klavs: No problem, good enough for now.

gql/lib/src/schema/definitions/value_types.dart
```diff
@@ -0,0 +1,198 @@
+// ignore_for_file: prefer_constructors_over_static_methods
```
klavs, line 1: Would be nice to get rid of this one as well.

PR #87, opened 2020-04-24
Catch IO errors

smkhalsa:
The `http.post()` call is currently outside the try / catch block, so IO errors are not getting caught and translated into LinkExceptions.


klavs:
Covered by #89

PR #89, opened 2020-04-24
Link exception improvements

klavs:
gql_link: cover Context access exceptions

gql_http_link: cover IO exceptions and context access exceptions


klavs:
@smkhalsa This covers #87 

I might have gone overboard with the exception usage here.

Any comments about this approach?

PR #95, opened 2020-05-02
multipack improvements

klavs:
Inspired by [GPM](https://github.com/dint-dev/gpm) I'm updating `multipack`.

I would love to use GPM, but it's lacking some key features of `multipack`.

See the diff of `.github/workflows/dart.yml` for side-by-side comparison of the old and new functonality.


klavs:
@smkhalsa, @micimize It would be great if you could do a test run of the new instructions in `README.md`. It worked fine for me, but I had some small issues on CI.

This MR will allow me to re-organize directories to put links in one sub-directory and examples in another.

@thosakwe I'm tagging you because I know you are working with a monorepo as well. Maybe this is interesting for you.

smkhalsa:
> @smkhalsa, @micimize It would be great if you could do a test run of the new instructions in `README.md`. It worked fine for me, but I had some small issues on CI.
> 
> This MR will allow me to re-organize directories to put links in one sub-directory and examples in another.
> 
> @thosakwe I'm tagging you because I know you are working with a monorepo as well. Maybe this is interesting for you.

Just tried the commands in the `README.md`, and it seems to be working correctly.

multipack/bin/multipack.dart
```diff
@@ -0,0 +1,50 @@
...
+
+  final runner = CommandRunner<void>(
+    "multipack",
+    "Manage monorepo.",
+  );
+
+  runner.argParser.addMultiOption(
+    "only",
+    abbr: "o",
+    help: "Whitelist packages.",
```
micimize, line 30:
```suggestion
    help: "Whitelist packages, skipping those not included for this command.",
```

multipack/bin/multipack.dart
```diff
@@ -0,0 +1,50 @@
...
+    defaultsTo: packageNames,
+    allowed: packageNames,
+  );
+
+  runner.argParser.addMultiOption(
+    "skip",
+    abbr: "s",
+    help: "Blacklist packages.",
```
micimize, line 38:
```suggestion
    help: "Blacklist packages for this command.",
```

README.md
```diff
@@ -127,11 +127,53 @@ Ideas for future work:
...
+to run commands in topological order.
+```text
+Available commands:
+  analyze   Run analyzer.
+  exec      Execute any command.
+  fmt       Run formatter.
+  pub       Run pub.
+  pubspec   Update pubspec.yaml
+``` 
+
```
micimize, line 150:
````suggestion

`pubspec` has 3 subcommands:
```
Available subcommands:
  clean           cleans dependency overrides
  hard_override   overrides dependencies for local packages
  override        overrides dependencies for local packages
```
````

micimize:
Oh just realized that `hard_override`'s links aren't reset by `clean`, and the `--help` is the same as for `override`

klavs:
There are definitely some dark corners. I'm thinking of rewriting the functionality of the `pubspec` command so it's not as polished.

> it looks like `flutter` vs `flutter pub` is automatically handled, which is awesome and should be documented

I took the idea from GPM. At this point I do not want to invest too much time in documentation beyond the basics necessary for contributing to gql.

> `multipack fmt --help` runs and prints help for each package – would be nice to handle

This one's a bit tricky because I'd like to avoid adding special cases of what is passed through `fmt` as-is and what is handled in a special way. I guess it'd be good to add a `help` command for this purpose. `multipack help fmt`. (Also, it seems that both `pub`s are not equal.)

> I like the colors, but a config for colors would be nice. They look weird in my solarized terminal:

This is a good one. Once there's a config file, color settings should be configurable there.

> Other than that, it's a bit awkward to me to have `multipack` in the same repo as `gql*`

True. The only reason is that it's a tool I added to manage this repo specifically (the same reason why it is not published). If other projects are interested in using this over GPM, we'll just move it out. At this point I do not see myself actively supporting other use cases.

PR #96, opened 2020-05-03
reorganize directories

klavs:


smkhalsa:
This is definitely easier to grok than the previous directory structure.

Issue #57, opened 2020-02-16
Bring Links up to par with what package:graphql has


klavs:
- [x] Example of Auth link functionality
- [x] Error Link
- [x] Websocket Link
- [x] File upload
- [x] http_link should have something like the useGETForQueries option as well
- [ ] persisted queries were recently contributed

DoctorJohn:
I wrote an auth link that works with the gql http link. It's similar to the one in the `graphql_flutter` package has. I may contribute mine, but where to put it? It's just 21 lines...

klavs:
Hey @DoctorJohn, welcome!

A new link should go to a new package at the root of this directory.

To kickstart your PR you can start by adding it to the example directory of `gql_httpl_link` so that we can discuss the design.

smkhalsa:
Here's the basic implementation of `ErrorLink` that I'm using in case it's helpful:

```dart
import 'dart:async';
import 'package:ferry/ferry.dart';
import 'package:gql_link/gql_link.dart';
import 'package:gql_exec/gql_exec.dart';

typedef void OnError({
  LinkException linkException,
  List<GraphQLError> graphQLErrors,
});

class ErrorLink extends Link {
  final OnError onError;

  ErrorLink(this.onError);

  @override
  Stream<Response> request(Request request, [forward]) =>
      forward(request).transform(
        StreamTransformer.fromHandlers(
          handleError: (error, stacktrace, sink) {
            if (error is LinkException) {
              onError(linkException: error);
            }
            sink.addError(error);
          },
          handleData: (data, sink) {
            if (data.errors != null && data.errors.isNotEmpty) {
              onError(graphQLErrors: data.errors);
            }
            sink.add(data);
          },
        ),
      );
}

```

micimize:
- [ ] `http_link` should have something like the [`useGETForQueries`](https://github.com/zino-app/graphql-flutter/blob/43ff4d394881c2f3175d3dd96ef7d372dc17e9a0/packages/graphql/lib/src/link/http/link_http.dart#L52) option as well
- [ ] [persisted queries](https://github.com/zino-app/graphql-flutter/blob/beta/packages/graphql/lib/src/link/apq/link_persisted_queries.dart) were recently contributed

klavs:
I think `useGETForQueries` wasn't doing anything. Am I wrong?

micimize:
[...]

PR #103, opened 2020-05-07
Implement `ErrorLink`

klavs:
- `ErrorLink` implementation
- An example `HttpAuthLink` using both `ErrorLink` and `TransformLink`


klavs:
@smkhalsa, this is a bit different than what you suggested in #57

The difference arises from a need to recover from errors silently if possible.

gql_error_link/lib/gql_error_link.dart
```diff
@@ -0,0 +1,68 @@
...
+  final ErrorHandler onError;
+  final ExceptionHandler onException;
+
+  ErrorLink({
+    this.onError,
+    this.onException,
+  });
+
+  @override
+  Stream<Response> request(Request request, [forward]) async* {
```
smkhalsa, line 37: I believe async generators in Dart terminate after receiving an error (which is why I opted for a StreamTransformer). I presume we want ErrorLink to be recoverable, especially when using a WebSocketLink as the terminating link.

klavs: You are most likely right about it!
I should write tests to verify that behavior.

I'll try [Result](https://pub.dev/documentation/async/latest/async/Result-class.html) stream. To me async generators are easier to write, read and understand (with the exception of the error behavior). If it works out well, I'm considering a breaking change in Links from `Stream<Response>` to `Stream<Result<Response>>`.

smkhalsa: Just to make sure I understand, are you suggesting that all errors would be passed in the Result, meaning that we wouldn't have to worry about code in the async generator throwing? 

I agree that async generators are easier to read and understand. However, I'm concerned about introducing an alternative mechanism for handling errors when Dart Streams already support both data and error events. Users implementing custom links wouldn't be able to rely on the built-in stream error handling.

klavs: > Just to make sure I understand, are you suggesting that all errors would be passed in the Result, meaning that we wouldn't have to worry about code in the async generator throwing?

Yes.

>  However, I'm concerned about introducing an alternative mechanism for handling errors when Dart Streams already support both data and error events.

I understand your concerns. `Result.captureStream` and `Result.releaseStream` allow easy transition between the std streams and `Result` streams.
`Result` from `package:async` which is pretty much the extended std lib of Dart. 

Anyway, I'm now writing tests so that I can compare `Result` stream vs std stream.

klavs: Take a look at this: https://github.com/gql-dart/gql/pull/103/commits/441823ff23343fcdc4fdc0557c46ac8341299921

I worked around the problem by using a `Result` stream. If `Result` was part of the Link API, the capture/release bookkeeping would not be necessary.

klavs: @smkhalsa, I've implemented more tests. Do you think there's a better way to write that link?

klavs: I finally stumbled upon a simple workaround to the generator problem:
```
  final generate = () async* {
    yield 1;
    yield* Stream<int>.error("a");
    yield* Stream<int>.error("b");
    yield 2;
  };
```

klavs: Apologies for the noise, I worked way too late to realize the edge cases.

I played around and realized that changing the Link API is not worth it as it becomes too complicated in other use cases.

I think the current approach where `Result` is used internally is the best way to go about it.

Also I aligned the implementation to `apollo-link-error`. So it's no longer possible to both return a new stream and continue the original.

gql_error_link/lib/gql_error_link.dart
```diff
@@ -0,0 +1,79 @@
...
+          final stream = onException(request, forward, error);
+
+          if (stream != null) {
+            yield* stream;
+
+            return;
+          }
+        }
+
+        yield* Stream.error(error);
```
smkhalsa, line 58: Will this still allow the link to receive data events after an error?

I'm still not clear on why we wouldn't just use StreamTransformers. It seems like async generators with Result is actually making it more complicated, not less.

klavs: > Will this still allow the link to receive data events after an error?

Yes. This yields an error without closing the generator.

> I'm still not clear on why we wouldn't just use StreamTransformers.

`StreamTransformer.fromHandlers` appears to be synchronous so it's not possible to redirect the stream and guarantee the `sink` will not be closed before all events are consumed. (Which is a requirement for AuthLink).
If we follow Apollo, Error Link is not a transformer at all. It's a way to get notified about errors and possibly re-sourcing the stream.

smkhalsa: Ok, that makes sense.

gql_error_link/lib/gql_error_link.dart
```diff
@@ -0,0 +1,79 @@
...
+
+/// [ErrorLink] allows interception of GraphQL errors (using [onError])
+/// and [LinkException]s (using [onException]).
+///
+/// In both cases [ErrorLink] transfers control over to the handler which may
+/// return a new stream to discard the original stream. If the handler returns
+/// `null`, the original stream is left intact and will be allowed to continue
+/// streaming new events.
+class ErrorLink extends Link {
+  final ErrorHandler onError;
```
smkhalsa, line 31: Do we want to call this onGraphQLErrors to make it explicitly clear that this is not an error in the link?

klavs: Yeah. Done.

PR #105, opened 2020-05-08
Avoid WSLink config null error

comigor:
Instantiate a default `SocketClientConfig` on `WSLink` to avoid "was called on null" errors.


comigor:
cc @iscriptology

Issue #119, opened 2020-06-15
[gql_http_link] Add response headers to context


micimize:
In `graphql-flutter` I closed an issue about a specific solution for accessing response headers a while ago, but never followed up with the correct solution (adding headers to context)

https://github.com/zino-app/graphql-flutter/issues/371

joshuadeguzman:
Hey @micimize,

I am using cookies for authenticating my requests. I found a [workaround](https://github.com/zino-app/graphql-flutter/issues/592)) at least for now for achieving it.

Curious about your thoughts in terms of performance passing a custom http client.

Thank you!

micimize:
@joshuadeguzman any performance hit would be due to the overall performance diff between `dio` and `http`, which I would expect to be at most negligible. Also `http` also has some serious long-standing open issues that do not inspire confidence

PR #114, opened 2020-05-30
Multipart File Upload and useGETForQueries support

micimize:
This extracts the logic for multipart file upload and `useGETForQueries` from `graphql/client.dart`, with accompanying tests

**Major changes**:
* `useGetForQueries` argument to `HttpLink`
* using `http.Client.send` instead of `post`
* `HttpLink` stores `uri` as a `Uri`
* I reworked `HttpLink._serializeRequest` into an exception handling HOF to handle the different paths, but really I think `RequestSerializer` should have a generic `Serialized` type which is `http.BaseRequest` for `HttpLink`. Otherwise it can't really handle the serialization concern IMO

This addresses some of #57

PR #121, opened 2020-06-18
[gql_http_link] Add response headers to context

agent3bood:
Related issue https://github.com/gql-dart/gql/issues/119


links/gql_http_link/lib/src/link.dart
```diff
@@ -32,13 +32,21 @@ class HttpLinkResponseContext extends ContextEntry {
...
 
+  /// HTTP response headers
+  final Map<String, String> headers;
+
   const HttpLinkResponseContext({
     @required this.statusCode,
-  }) : assert(statusCode != null);
+    @required this.headers,
+  }) : assert(
+          statusCode != null,
```
joshuadeguzman, line 42: Thanks for this. Maybe we can add some explicit error messages here?

links/gql_http_link/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.2.10
+
+- add `headers` to `HttpLinkResponseContext`
```
joshuadeguzman, line 3: I still think that this should be included on a release commit.

agent3bood: What do you mean, should I remove this change?

links/gql_http_link/lib/src/link.dart
agent3bood: I have followed the same assert style on the project, split it in two lines and keep it without error message.

links/gql_http_link/CHANGELOG.md
micimize: My understanding is we [update the changelog in the change PR](https://github.com/gql-dart/gql/pull/114#issuecomment-640757142), so this is right. @klavs seems we should probably make a more detailed contributor doc.

links/gql_http_link/lib/src/link.dart
micimize: Right now this checks `statusCode != null` with the message `headers != null` should be:
```dart
assert(statusCode != null),
assert(headers != null);
```
The messages here would be too generic to be worth including IMO

agent3bood: I fixed this mistake.

agent3bood:
@micimize What should I do now? 
I am not sure if I am required to do something or if we are waiting for @klavs

micimize:
@agent3bood it LGTM so yeah it's just waiting on @klavs

klavs:
> My understanding is we update the changelog in the change PR, so this is right. @klavs seems we should probably make a more detailed contributor doc.

@micimize, agreed. I'm not in favor of the Angular-style commit messages. That's been a major blocker for me offering small fixes to projects like graphql_flutter. I never liked spending more time reading the policies than coding the fix.

I prefer being open to "free-style" contributions with automatic checks (as much as that is possible)

PR #122, opened 2020-06-28
Apply the async initialPayload handling logic

RodolfoSilva:
As described here (https://github.com/zino-app/graphql-flutter/issues/658), we need to apply the async initialPayload handling logic.


RodolfoSilva:
Thanks @micimize

PR #123, opened 2020-07-26
context-configured GET requests and serialization config

micimize:
This is a draft of functionality I added to make `PersistedQueriesLink` work in `graphql_flutter` v4:
* Adds a `HttpLinkMethod` for context-based GET requests
* Adds a `RequestSerializationInclusions` for configuring serialization

I'm pretty unsure about both APIs, but essentially `PersistedQueriesLink` wants to set `GET` in context and have a way to set `includeQuery: false`.

### DIFF
--- a/links/gql_http_link/lib/src/link.dart
+++ b/links/gql_http_link/lib/src/link.dart
@@ -27,6 +27,29 @@ class HttpLinkHeaders extends ContextEntry {
       ];
 }
 
+/// HTTP link request method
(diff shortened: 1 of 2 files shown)


brightchip:
Please how do I enable persisted queries ? 
There's no PersistedQueriesLink in the   graphql_flutter: ^5.0.1-beta.1. 
This is a must feature for my project. How do I use it

PR #126, opened 2020-08-02
[codegen] Add built_value var and data classes for fragments

smkhalsa:
This PR adds support for generating variable classes for fragments.

Since variable types are not available in fragment definitions the way they are in queries, we recursively traverse the fragment, and when a variable is found, we look up the argument type from the schema.

This PR also adds built_value data classes for fragments.


smkhalsa (commented):
Two checks are failing due to a difference in ordering of generated code on each build which is getting caught by the git diff check. 

This should be OK to merge regardless.

Issue #115, opened 2020-06-04
[gql_link] exceptions


agent3bood:
I am working on adding exceptions to `WebSocketLink`, the plan is 
1. `WebSocketLinkParserException` Exception occurring when parsing fails or parsed response is missing both `data` and `errors`.

2. `WebSocketLinkServerException` Exception occurring when network fails.

Right now the error "parsed response is missing both `data` and `errors`." is handled as [`ServerException`](https://github.com/gql-dart/gql/blob/master/links/gql_link/lib/src/exceptions.dart#L54)

I want to refactor this as [`ResponseFormatException`](https://github.com/gql-dart/gql/blob/master/links/gql_link/lib/src/exceptions.dart#L31).

Any suggestions, should I continue as planned?

agent3bood:
Code review and comments are welcome, I will write tests once we agree on this.

PR #130, opened 2020-08-15
websocket link exceptions 

agent3bood:
Related issue https://github.com/gql-dart/gql/issues/115
All exceptions have the same behavior of HttpLink exceptions


klavs:
Please update Changelog and

PR #131, opened 2020-08-16
[codegen] Generate variable classes for fragments

smkhalsa:
This PR adds support for generating variable classes for fragments.

Since variable types are not available in fragment definitions the way they are in queries, we recursively traverse the fragment, and when a variable is found, we look up the argument type from the schema.


smkhalsa:
closed in favor of https://github.com/gql-dart/gql/pull/126

PR #145, opened 2020-09-26
[gql_exec] Print operation

micimize:
errors are hard to parse without having the actual operation printed.

@klavs On the other hand, maybe that's ok/desired so that the output is legible?


klavs:
No strict opinions here. Looks good.

Issue #154, opened 2020-10-12
[discussion] Change HttpLink interface to make client custom type


agent3bood:
This discussion might be late to open, I have seen an open PR for Dio-link https://github.com/gql-dart/gql/pull/150
I want to make a suggestion and see if it is possible to implement.

instead of creating a new link for different http packages, we can change the interface of `HttpLink` and change the `_httpClient`  type.

The suggested `HttpLink` will look like
```Dart
  HttpLink(
    String uri, {
    this.defaultHeaders = const {},
    this.useGETForQueries = false,
    GqlDartHttpClient httpClient,
    this.serializer = const RequestSerializer(),
    this.parser = const ResponseParser(),
  }) : uri = Uri.parse(uri) {
    _httpClient = httpClient ?? http.Client();
  }
```

```Dart
abstract class GqlDartHttpClient {
  Response send(Request request);
}
```

This way it will be easy to implement any link, remove duplicate code which might introduce inconsistency.

Best
Abdullah

klavs:
Let's finalize the dio merge in the coming days and then work on abstractions to make both links inter-changeable.

agent3bood:
Thanks @klavs I will keep this open until the Dio PR is merged.

TarekkMA:
Yes this is the best approach I was thinking of making this because I was just keeping up with the gql_http_link features

klavs:
#150 is now merged in.
@agent3bood, @TarekkMA you have a green light. My design goal would be to either make them interchangeable or merge them in a single link using something like what @agent3bood suggested.

agent3bood:
Hi @TarekkMA your suggestion is perfect. Do you have time to finish it?

Check this example 
```Dart
class DioClient extends GqlDartHttpClient  {
  DioClient(Dio this.dio);
  final Dio dio;

  @override
  Response send(Request request) {
    // TODO
  }
}
```

agent3bood:
Closing reason https://github.com/gql-dart/gql/pull/157#issuecomment-817144191

PR #157, opened 2020-10-21
Extract http client from http link, allowing custom clients (dio)

TarekkMA:
This WIP-PR that aims to solve the issue discussed in #154.

The old API will remain almost the same. `HttpLink` default constructor will create a new `HttpGqlClient`. The `HttpLinkParserException` & `HttpLinkServerException` will now take the abstracted `GqlClientResponse` type instead of the `http` lib type.

All tests are passing, but the test that depend on the  `http.Response` type in the exceptions.

Suggestions for how to improve the new API are more than welcome.


links/gql_http_link/lib/src/gql_client.dart
```diff
@@ -0,0 +1,110 @@
...
+        o.method == method &&
+        o.uri == uri &&
+        mapEquals(o.body, body) &&
+        mapEquals(o.headers, headers) &&
+        mapEquals(o.files, files) &&
+        mapEquals(o.fileMapping, fileMapping);
+  }
+
+  @override
+  int get hashCode =>
```
micimize, line 69: What about inheriting from `gql_exec.Request/Response` and use the same [==/hashCode](https://github.com/gql-dart/gql/blob/7b1dd8a889fff670e6506c020afe2d3fd5866418/links/gql_exec/lib/src/request.dart#L42-L64) pattern we take there?

micimize (commented):
I like this approach overall. What other file types are you considering aside from `MultipartFile`?

TarekkMA:
> I like this approach overall. What other file types are you considering aside from `MultipartFile`?

Do we need to support other types? We can have the GqlClient.checkFileType include things like dart:io.Files and let each client implementation handle it.

links/gql_http_link/lib/src/gql_client.dart
TarekkMA: The ==/hashCode pattern there is cleaner.

and how will be handle RequestSerializer and ResponseParser that is included with the link.

I was thinking of it as an apstraction layer above the Request/Response of each http client. inheriting from gql_exec.Request/Response will need the serializer and parser to be also a part of the object I think.

agent3bood:
@TarekkMA Sorry I was busy and could not make a full review.
I have one question, is it possible get rid of `GqlClientRequest` and `GqlClientResponse` and use the existing `Request` and `Response` classes?
What are the reasons that made you make these two classes?

TarekkMA:
> @TarekkMA Sorry I was busy and could not make a full review.
> I have one question, is it possible get rid of `GqlClientRequest` and `GqlClientResponse` and use the existing `Request` and `Response` classes?
> What are the reasons that made you make these two classes?

The reason I made those 2 classes is to abstract the http request/response for each client. So you will find `headers` and `filesMap`, etc. and the response will have things like `statusCode`.

AFAIK `gql_exec.Request/Response` don't serve this purpose, and I didn't want to add extra functionally to these classes.

So right now the client have these responsibilities:
- Execute a request based on the data found in the passed `GqlClientRequest` object.
- Map the http response to `GqlClientResponse` object and return it.

And the link responsibilities are:
- Serialize the request based on the `RequestSerializer` passed to the link.
- Convert the serialized request to `GqlClientRequest` object and configure it based on parameters passed to the link and files attached to the request if found.
- Send the `GqlClientRequest` object to be executed from the client and receive an `GqlClientResponse` object
- Parse the returned `GqlClientResponse` object using `ResponseParser` and return that.

Sorry if I missed anything. Feel free to suggest alternatives to what I proposed.

agent3bood:
I think all the logic that is currently in `GqlClientRequest` and `GqlClientResponse` are related to httpLib.client and therfore it should be in the custom `HttpGqlClient`, but this may require modifying the current `Request` & `Response` and library API.
Do you think it is possible?

TarekkMA:
@agent3bood no the objects is shared between the `dio` and `http` lib. Take a look at my implantation of the `gql_dio_link` (found in this PR) which is a wrapper for the `gql_http_link` with custom client ,`DioGqlClient`,.

Currently, the `gql_dio_link` passes the old tests and it does now have file upload feature, the logic of which is abstracted in the `GqlClientRequest ` object,.

links/gql_http_link/lib/src/gql_client.dart
```diff
@@ -0,0 +1,114 @@
...
+      method: method,
+      uri: uri,
+      body: body,
+      headers: headers,
+      fileMapping: fileMapping,
+      files: files,
+    );
+  }
+
+  bool get isMultipartRequest => files != null && files.isNotEmpty;
```
agent3bood, line 52: can be simplified `files?.isNotEmpty == true`

links/gql_http_link/lib/src/gql_client.dart
```diff
@@ -0,0 +1,114 @@
...
+
+class GqlClientRequest {
+  final String method;
+  final Uri uri;
+  final Map<String, dynamic> body;
+  final Map<String, String> headers;
+  final Map<String, dynamic> files;
+  final Map<String, List<String>> fileMapping;
+
+  GqlClientRequest({
```
agent3bood, line 11: Mart the required fields as `@required` or add default values
method, uri, body, headers as required  (not sure about file and fileMapping)

links/gql_http_link/lib/src/gql_client.dart
```diff
@@ -0,0 +1,114 @@
...
+    for (int i = 0; i < fileMapEntries.length; i++) {
+      final MapEntry<String, dynamic> entry = fileMapEntries[i];
+      final String indexString = i.toString();
+      fileMapping.addAll(<String, List<String>>{
+        indexString: <String>[entry.key],
+      });
+      files[indexString] = entry.value;
+    }
+
+    return GqlClientRequest(
```
agent3bood, line 42: Mart the required fields as `@required` 
(I assume all fields are required)

links/gql_http_link/lib/src/gql_client.dart
```diff
@@ -0,0 +1,114 @@
...
+        DeepCollectionEquality(),
+      ).hash(_props);
+}
+
+class GqlClientResponse {
+  final int statusCode;
+  final Map<String, String> headers;
+  final Map<String, dynamic> body;
+
+  GqlClientResponse({
```
agent3bood, line 82: Mart the required fields as @required
(I assume all fields are required)

agent3bood (commented):
LGTM

links/gql_http_link/lib/src/gql_client.dart
TarekkMA: Yeah, much cleaner

TarekkMA:
I'm still not sure if this API is good enough and whether we should extract the client from the `gql_http_link` library to be its own library, maybe `gql_http_client`. So dio and potential other clients don't pull extra dependencies.

agent3bood:
You are right, it com, complicates the api, the final say will be to @klavs

klavs:
@TarekkMA, @agent3bood please finalise in whichever way it makes sense to you. You'll always be able to improve it later.

TarekkMA:
@agent3bood, @klavs sorry about the long absence, had some exams and couldn't work on this PR.

Can you check what is missing from this PR to be merged?

I think we will make `gql_dio_link` depend on `gql_http_link` for now, and will extract the `gql_http_clinet` in its own package later.

kateile:
Can someone please merge this?

agent3bood:
This PR has made the API more complicated instead of simplifying it, I don't think it will be merged now.
However it is easy to implement custom link, just copy the `gql_http_link` and start from there.

syssam:
What happened?

I make a pr for a long time ago.
https://github.com/TarekkMA/gql_dio_link/pull/2/files

We just follow the gql_http_link implement to enhnace gql_dio_link

TarekkMA:
I'm very sorry for that, I'm currently not available these couple of months. Very sorry about that 😟

agent3bood:
@TarekkMA I will close this PR, it has taken too long to come up with a solution. Another issue is our current solution did complicate the package API and added a tight coupling.

Let the link implementer write full functionality, there aren't many links to be implemented anyways.

PR #158, opened 2020-10-22
Fix triplicated content-type headers on gql_dio_link

Grohden:
I've fixed an issue with the package that was adding twice the content-type header, thus causing the request to fail for graphql.

There's a discussion here for further details:
https://github.com/TarekkMA/gql_dio_link/issues/4


TarekkMA (approved):

TarekkMA:
I think you can merge now.

Thanks for the PR.

Grohden:
> I think you can merge now.

@TarekkMA I think I don't have permissions to merge here XD
Not sure who could merge it tho

PR #162, opened 2020-11-08
[gql_websocket_link] Add onDisconnect callback + discussion on websocket link lifecycle [WIP]

iscriptology:
As I was working with the websocket link I ran into the situation where my code was not aware of the client closing itself because of a network cutoff or some other reason.
I had to add a callback for disconnection so I can re-create the link for my dependencies tree. (this is the PR, I don't expect it to be merged, just to create a discussion)

The best solution I think would be to have some ReconnectingWebSocketLink that will do it within itself, but it will have to re-send the previously sent subscriptions to the server or otherwise they will stop arriving. (because the server sees the reconnection as a brand new client/socket) fix me if I'm wrong here.

I went through [this](https://github.com/enisdenjo/graphql-ws/blob/master/src/client.ts) implementation and I didn't see anything regarding resending the subscriptions. (this code is not so readable and full of closures that return closures so I'm not confident that I understood it fully)

Maybe resending the subscriptions has implications that I'm not aware of. in any case there must be a better way for the library's user to track the lifecycle of the link and act accordingly.

Anyone with an implementation proposal? a better approach?

### DIFF
--- a/links/gql_websocket_link/lib/src/link.dart
+++ b/links/gql_websocket_link/lib/src/link.dart
@@ -52,8 +52,7 @@ class WebSocketLink extends Link {
   static const int open = 1;
   static const int closing = 2;
   static const int closed = 3;
-  final BehaviorSubject<int> _connectionStateController =
-      BehaviorSubject<int>();
+  final BehaviorSubject<int> _connectionStateController = BehaviorSubject<int>();
 
   Stream<GraphQLSocketMessage> _messageStream;
   StreamSubscription<GraphQLSocketMessage> _messageSubscription;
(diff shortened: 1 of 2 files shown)


iscriptology:
Actually digging in a little deeper it seems like every subscription [there](https://github.com/enisdenjo/graphql-ws/blob/master/src/client.ts) is sent in an infinite loop until canceled by the user. Which is actually great cause the subscriptions are resent on reconnection.

We need an implementation like this here I believe.

agent3bood:
Thanks for sharing you code, I am having the same issue. I will share my implementation once I have something worthy.

agent3bood:
Once the socket is closed, the client cannot receive the subscriptions again, it has to send a new subscription request.
Check here https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/client.ts#L510

I have initial working code, I try to publish it once it is cleaned up and checked.

ram231:
I would like to participate in this discussion. But I have no prior knowledge on how subscription works under the hood.

I had a similar case where the subscription suddenly disconnect and either the `onDone` or the `onError` doesn't call in my Stream.

agent3bood:
I am closing this issue, the websocket link has support for subscription with auto reconnect now.

Issue #176, opened 2020-12-07
[gql_http_link] allow custom _parseHttpResponse function


agent3bood:
### Background

It is recommended to parse json in background 
https://flutter.dev/docs/cookbook/networking/background-parsing

### Issue

Right now the `gql_http_link` has a function `_parseHttpResponse` which does parse the json on the same thread as the UI thread.

### Solution

Allow passing a custom `_parseHttpResponse` 

I am working on this and will submit a PR soon.

BaudouinCS:
Hello @agent3bood 

I have great troubles try/catching these `HttpLinkParserException` thrown from `_parseHttpResponse`.

Would you mind pointing out how to catch them for me please? Would be greatly appreciated!

Regards.

TiffApps:
Having the same issue, would be nice to have an example

knaeckeKami:
can you give more details?

BaudouinCS:
@knaeckeKami See [this issue](https://github.com/heftapp/graphql_codegen/issues/214) for more details, they redirected me to this package.

If it doesn't belong to the current issue, I can open a new one for clarity.

BaudouinCS:
@knaeckeKami Hello, have you got the opportunity to look into this by any chance?

knaeckeKami:
the link thows an exception on status >= 300. The client that you use will probably wrap the exception in the response object.

PR #177, opened 2020-12-07
custom http response decoder

agent3bood:
Related issue https://github.com/gql-dart/gql/issues/176


links/gql_http_link/lib/src/link.dart
```diff
@@ -9,6 +9,9 @@ import "package:meta/meta.dart";
 import "./_utils.dart";
 import "./exceptions.dart";
 
+typedef HttpResponseDecoder = Future<Map<String, dynamic>> Function(
```
knaeckeKami, line 12: could we change this to `FutureOr<Map<String, dynamic>>` ? then the default encoder could still be sync

agent3bood: Return type changed.

Thanks

klavs (approved):

PR #181, opened 2020-12-22
Draft: gql NNBD

klavs:


smkhalsa:
@klavs I know you're stepping back from this project, but since this PR is still open, I'm wondering if you have any further thoughts about implementing NNBD in the `gql` packages.

Are you satisfied with the lexer / parser conversion that you've already implemented in this PR? Any other thoughts/suggestions for pushing this forward?

klavs:
Work on NNBD in `package:gql` should definitely be done. I've received email from Dart team about `package:gql` as "a package that we've identified as popular and important to our ecosystem". It makes sense to transition as we are kind of blocking others from transitioning as the `package:gql` is very low level tool used by most of the GraphQL ecosystem in Dart.

It's a bit of an exercise to adjust to NNBD at the same time of reintroducing myself to that code. I think I actually caught one bug in this CI already.

PR #183, opened 2021-01-10
CDN redirect fix (status code 301 & 302)

amin0075:
my domain is behind a CDN( Cloudflare ) and every time I request to the domain my request status code will be 301.so I added a new  loop for http request if the response status is 301 or 302 for handling redirects. the goal is if the mentioned status codes occurred try to request with the new location that exists in the header of response. this method will recursively request to the new location till the status code changes to something else. if its not a way to handle redirects please let me know about it.

I followed this solution :  https://github.com/zino-app/graphql-flutter/issues/576#issuecomment-626246953

### DIFF
--- a/links/gql_http_link/lib/src/link.dart
+++ b/links/gql_http_link/lib/src/link.dart
@@ -62,7 +62,7 @@ class HttpLinkResponseContext extends ContextEntry {
 /// [http.Client] to the constructor.
 class HttpLink extends Link {
   /// Endpoint of the GraphQL service
-  final Uri uri;
+  Uri uri;
 
   /// Default HTTP headers
   final Map<String, String> defaultHeaders;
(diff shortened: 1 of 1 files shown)


agent3bood:
Thanks for the contribution, ion theory the link will [follow redirects](https://pub.dev/documentation/http/latest/http/BaseRequest/followRedirects.html)

What is the link response? if possible paste it here.

can you try replacing the following code in https://github.com/gql-dart/gql/blob/master/links/gql_http_link/lib/src/link.dart#L225 ?

```
    final request = http.Request("POST", uri)
      ..body = httpBody
      ..headers.addAll(headers);
    request.followRedirects = true;
    return request;
```

amin0075:
thanks, I changed my URL from http to https and it worked fine, so it won't redirect from http to https.
btw redirection works in browser.

PR #187, opened 2021-02-09
gql_websocket_link: Delay requests until connect is done

janosroden:
If `initialPayload` is an async function then wait for it before send any requests.
For example acquire a JWT token.


janosroden:
@agent3bood Could you please take a look?

agent3bood:
Will check it.

Why are there changes to `gql.svg` there is another PR with an issue with the svg. Did you modify it or is it auto generated?

janosroden:
It's validated by CI build (I copied it from the logs)

agent3bood:
The code should await for the websocket to open before sending any request.

Can you try another way of fixing the issue you are facing.

1.
Remove the `.startWith(open)` from here `waitForConnectedState`

```
      final Stream<int> waitForConnectedState = _connectionStateController
          .where((state) => state == open)
          .take(1);
```

https://github.com/gql-dart/gql/blob/master/links/gql_websocket_link/lib/src/link.dart#L113

2.
in `_connect()`

move the `_connectionStateController.value = open;` to after sending the initial payload 

```
if (initialPayload is Function) {
        final dynamic payload = await initialPayload();
        _write(InitOperation(payload));
      } else {
        _write(InitOperation(initialPayload));
      }
_connectionStateController.value = open;
```

agent3bood:
@janosroden  Did my solution fix the issue for you?

janosroden:
@agent3bood Yes, I just tried, your solution works perfectly. Thanks.

agent3bood (approved):
LGTM

PR #193, opened 2021-02-13
Nullsafe gql, gql_exec, gql_link

micimize:
This is a first pass at a nullsafe `gql`, mostly done with `dart migrate` and applied quick fixes. Idk what the status of `nnbd-gql` is – this was basically a "follow the instructions and see if tests pass" approach.


micimize:
Null safety migrations run thus far:
 - [X] gql
 - [ ] gql_build
 - [ ] gql_code_builder
 - [X] gql_exec
 - [X] gql_link
 - [ ] gql_dedupe_link
 - [ ] gql_dio_link
 - [ ] gql_error_link
 - [ ] gql_http_link
 - [ ] gql_websocket_link
 - [ ] gql_transform_link
 - [X] cats
 - [ ] gql_example_http_auth_link

klavs:
Great work!
I can see this deployed as the initial nnbd version.
But before the final nnbd version is deployed, gql AST must be rechecked manually with GraphQL spec at hand. I know there are some fields on AST classes which are non-nullable and some are nullable.

I kind of forgot about the `dart migrate` when I did my initial work on nnbd (in some other branch)

micimize:
migrated `gql_http_link` – discovered two things:
* We'll have to wait on `build_runner` for the build packages #2920 
* Tests with mockito will have to add some [code gen](https://github.com/dart-lang/mockito/blob/master/NULL_SAFETY_README.md), and not add the autogenerated `!` assertions in tests.

I'm thinking I should actually merge this into a `gql-dart:nullsafety` branch here and publish `-nullsafety.alpha.0` versions manually as they're ready

klavs:
> I'm thinking I should actually merge this into a gql-dart:nullsafety branch here and publish -nullsafety.alpha.0 versions manually as they're ready

I thought about the same thing because of the auto-release.

gql/lib/src/language/lexer.dart
```diff
@@ -378,10 +378,10 @@ class _Scanner {
           case 116: // \t
             break;
           case 117: // \u
-            final isUnicode = isHex(peek(offset: 1)) &&
-                isHex(peek(offset: 2)) &&
-                isHex(peek(offset: 3)) &&
-                isHex(peek(offset: 4));
+            final isUnicode = isHex(peek(offset: 1)!) &&
```
klavs, line 381: It's actually a bug to use `!` operator here. (And it has been a bug before as well)

https://github.com/gql-dart/gql/pull/181/files#diff-6864373280eabf861af843f08282de1328566600ca3df35cd073908df9745f6fR381

gql/lib/src/language/parser.dart
```diff
@@ -51,21 +51,21 @@ class _Parser {
     ++_position;
   }
 
-  Token _expectToken(TokenKind kind, [String errorMessage]) {
-    final next = _next();
+  Token _expectToken(TokenKind kind, [String? errorMessage]) {
+    final next = _next()!;
```
klavs, line 55: This is probably a bug as well.

gql/lib/src/language/printer.dart
```diff
@@ -583,13 +570,13 @@ class _PrintVisitor extends Visitor<String> {
   @override
   String visitDirectiveDefinitionNode(DirectiveDefinitionNode node) => [
         if (node.description != null) ...[
-          node.description.accept(this),
+          node.description!.accept(this),
```
klavs, line 573: I do not know NNBD that well, but this should be inferred by the if.

klavs (commented):
Almost every single use of `!` operator is a potential for a bug :(

gql/lib/src/ast/transformer.dart
```diff
@@ -379,7 +376,7 @@ class _Transformer extends Visitor<Node> {
     DirectiveDefinitionNode node,
   ) {
     final updatedNode = DirectiveDefinitionNode(
-      name: _visitOne(node.name),
+      name: _visitOne(node.name!),
```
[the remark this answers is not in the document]
micimize: The migration tool generated `N _visitOne<N extends Node?>(N node)`, which is a required non-null argument that extends a nullable argument. Added a `_visitNullable`, but there are a lot of arguments that look like `required NameNode? name` that are all suspect.

How did you resolve the class taxonomy initially? The spec assumes the availability of a type union operator which makes it very hard to map

.github/check_package.sh
```diff
@@ -0,0 +1,43 @@
+#/bin/bash
```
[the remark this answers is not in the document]
micimize: Added a note:
```bash
# NOTE:
#  This is a temporary utility script to help mitigate the need for github action
#  troubleshooting, and will be deleted either pre-merge or when melos replaces it
#  (https://github.com/gql-dart/gql/pull/192)
```

gql/lib/src/language/printer.dart
micimize: It can't, probably because getting `node.description` may somehow alter its next invocation.

dereferenced the attribute to avoid `!`

Issue #200, opened 2021-03-18
Package links broken in README


shyndman:
Hi there,

A number of the package links in the README are broken. Their intended destinations are easy enough to find, but I thought I should mention.

For example
* `package:gql_code_builder`
* `package:gql_code_build`
* `package:gql_exec`
* ...and so on. The entire set need a check.

agent3bood:
Thank for pointing this out, sometimes a nudge is needed to do simple fixes.

shyndman:
:) For sure

PR #201, opened 2021-03-18
fix links on README

agent3bood:
In response to issue https://github.com/gql-dart/gql/issues/200


agent3bood:
The `gql` depends on `cat` using a local path and not version  https://github.com/gql-dart/gql/blob/master/gql/pubspec.yaml#L14

Which is causing `multipack` to install `yaml 3.1.0` insteade of `yaml 2` which in return break the tests.

@klavs @micimize should the version of `cat` be fixed int the specified file?

PR #202, opened 2021-03-23
add null safety to codegen libs

smkhalsa:
- [x] update codegen libs to null safety
- [x] make codegen libs generate null-safe code

A few notes:

## `code_builder`

`code_builder` (upon which `gql_build` and `gql_code_builder` depend) has [not yet been migrated](https://github.com/dart-lang/code_builder/issues/303), and it's not clear when it will be, so I've migrated our codegen packages anyway. This means that the builders may have some unsound code paths for the time being, but this shouldn't really be an issue for users of the code gen.

## nulls in lists

The migration to null safety exposed a bug with nullable GraphQL types in lists. `BuiltList` doesn't support nulls (https://github.com/google/built_collection.dart/issues/232, https://github.com/google/built_value.dart/issues/1011), so nulls in lists will cause deserialization to fail. For example, let's take a look at the following schema:

```graphql
type User {
  name: String
  aliases: [String]
}
```
In this case, the `aliases` field will be a nullable list with nullable strings, which should be deserialized into the dart type `BuiltList<String?>?`. Since `BuiltList` cannot currently accept nulls, this fails.

I've implemented a [workaround](https://github.com/gql-dart/gql/blob/d34762ceb28654d58272a22b503157b3ab5ee18a/codegen/gql_code_builder/lib/src/common.dart#L99) to essentially ignore the nullability of values in lists such that the above example would be deserialized as a `BuiltList<String>?`. However, **if your graphql server actually returns a list with null values, it could cause a runtime exception.**

Since this issue already exists in the current non null-safe version, I'm simply highlighting it here. Once https://github.com/google/built_value.dart/issues/1011 is resolved, we will remove the workaround.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -21,7 +21,7 @@ jobs:
           - cats
     runs-on: ubuntu-latest
     container:
-      image: google/dart:2.12-beta
+      image: google/dart
     name: Check ${{ matrix.package }}
     env:
       PACKAGE: ${{ matrix.package }}
@@ -73,7 +73,7 @@ jobs:
           # gql_example_flutter would require flutter
     runs-on: ubuntu-latest
     container:
-      image: google/dart:2.12-beta
+      image: google/dart
     name: Check ${{ matrix.package }}
     env:
       PACKAGE: ${{ matrix.package }}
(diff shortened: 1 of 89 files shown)


smkhalsa:
@micimize I think this is pretty much ready to go. Any thoughts before merging into the main null safety PR?

micimize (approved):

micimize:
@smkhalsa looks fine to me! Want to go ahead and merge and we can sort out the build checks later?

smkhalsa:
@micimize many of the failing build checks were related to mismatch dart versions. I've updated the CI to use the latest 2.12.0 GA and updated the pubspecs accordingly.

PR #203, opened 2021-03-23
DioLink null safety

agent3bood:


micimize (approved):
I'm a bit on the fence with checking in the generated `.mocks.dart`. I avoided doing so myself, but that makes it so we have to run build runner just for unit tests which is annoying.

Approving and merging for now – can always delete later

PR #205, opened 2021-03-26
Add a method to stop subscriptions.

euriasb:
Currently there isn't any way to stop a subscription without closing the link altogether, this is problematic if you have multiple subscriptions and you want to be able to start and stop them without opening multiple sockets.

This PR adds an API to do this by exposing a 'stop' function on WebSocketLink, which takes the requestId.
The requestId was also not publicly available, so I'm also adding the RequestId context entry to the Response, it is up to the consumer to get the id from the response and use it to stop a subscription.

I added a unit test for this and I also tested it in my app which is under development.

### DIFF
--- a/links/gql_websocket_link/lib/src/link.dart
+++ b/links/gql_websocket_link/lib/src/link.dart
@@ -209,9 +209,9 @@ class WebSocketLink extends Link {
   /// ```
   Future<void> stop(String requestId) async {
     final messagesSubscription = _messageSubscriptions[requestId];
-    messagesSubscription?.cancel();
+    await messagesSubscription?.cancel();
     _messageSubscriptions.remove(requestId);
-    _write(StopOperation(requestId));
+    await _write(StopOperation(requestId));
     _requests.removeWhere((e) => e.context.entry<RequestId>().id == requestId);
   }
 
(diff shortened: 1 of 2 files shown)


agent3bood:
Hi @euriasb 

Thanks for contributing here. 

I think [this is](https://github.com/gql-dart/gql/blob/master/links/gql_websocket_link/lib/src/link.dart#L191) what you are looking for, which is available. If the feature is not working then we need to fix it instead of adding new method to the package.

Or is there something this PR offer that is not available right now? 

Also there is a test for it https://github.com/gql-dart/gql/blob/master/links/gql_websocket_link/test/gql_websocket_link_test.dart#L724

euriasb:
1. `_write` is not a public API, so it is not possible to send a call `_write(StopSubscription(id))`.
2. The subscription ID is also not publicly available, is only kept [here](https://github.com/gql-dart/gql/blob/master/links/gql_websocket_link/lib/src/link.dart#L133).

So this PR makes the subscription ID publicly available, and exposes a function to stop a subscription using that ID.

agent3bood:
The main objective of this PR is this statement `Currently there isn't any way to stop a subscription without closing the link altogether` which is not true, you can stop a subscription by stop listening on the stream. 


```
const myLink = WebSocketLink();
final sub = myLink.request(request);
// do work here
sub.cancel(); // this will send the STOP command
// you can still use the link after
```

euriasb:
Oh wow, I completely missed that, yes I just tried that and it worked, thanks for your help, I'm closing this PR now.

agent3bood:
You are most welcome.

PR #194, opened 2021-02-15
Null Safety

micimize:
Supersedes #193 as the nullsafety tracking PR. Will remain a draft and reference point for prereleases until dart has a stable null safe release. @klavs am planning on publishing `*-nullsafety.0` versions for migrated packages soon, so that downstream packages can begin migrating.

Null safety migrations run thus far:

### Libraries
 - [X] gql:                [`0.13.0-nullsafety.1`](https://pub.dev/packages/gql/versions/0.13.0-nullsafety.1)
 - [X] gql_exec:           [`0.3.0-nullsafety.1`](https://pub.dev/packages/gql_exec/versions/0.3.0-nullsafety.1)
 - [X] gql_link:           [`0.4.0-nullsafety.1`](https://pub.dev/packages/gql_link/versions/0.4.0-nullsafety.1)
 - [x] gql_http_link:      [`0.4.0-nullsafety.1`](https://pub.dev/packages/gql_http_link/versions/0.4.0-nullsafety.1)
 - [x] gql_dedupe_link:    [`2.0.0-nullsafety.1`](https://pub.dev/packages/gql_dedupe_link/versions/2.0.0-nullsafety.1)
 - [x] gql_error_link:     [`0.2.0-nullsafety.1`](https://pub.dev/packages/gql_error_link/versions/0.2.0-nullsafety.1)
 - [x] gql_transform_link: [`0.2.0-nullsafety.1`](https://pub.dev/packages/gql_transform_link/versions/0.2.0-nullsafety.1)
 - [x] gql_websocket_link: [`0.3.0-nullsafety.1`](https://pub.dev/packages/gql_websocket_link/versions/0.3.0-nullsafety.1)
 - [x] gql_dio_link: [`0.1.0-nullsafety.1`](https://pub.dev/packages/gql_dio_link/versions/0.1.0-nullsafety.1)

#### ~~blocked by https://github.com/dart-lang/build/issues/2920~~
 - [x] gql_code_builder
 - [x] gql_build

### Examples
 - [X] cats
 - [x] gql_example_cli
 - [x] gql_example_cli_github
 - [x] gql_example_build
 - [x] gql_example_http_auth_link
 - [x] gql_example_dio_link
- [x] gql_example_flutter

### stable release TODOs
- [x] ~~Hand-verify `gql` ASTs against the [current spec](https://spec.graphql.org/June2018/)~~

PR #208, opened 2021-04-01
Nullsafety gql example cli

agent3bood:


agent3bood:
Related https://github.com/gql-dart/gql/pull/194

PR #209, opened 2021-04-01
gql_dio_link: Forward original exception on dio errors

jamonkko:
Pass original exception also if http request fails with error status. Dio throws DioError if response status is in error range so this makes it possible to have that as original error and makes debugging easier.

### DIFF
--- a/links/gql_dio_link/CHANGELOG.md
+++ b/links/gql_dio_link/CHANGELOG.md
@@ -1,3 +1,3 @@
-## [0.0.1] - TODO: Add release date.
+## 0.0.5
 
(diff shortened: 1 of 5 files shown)


agent3bood (commented):
LGTM

agent3bood:
Thanks for contributing to this library.

Please bump the package version and add a line in CHANGELOG

jamonkko:
Version and changelog updated.

agent3bood (approved):

jamonkko:
@agent3bood Thank you (and all other maintainers) for doing this work! Really appreciate this library.

PR #210, opened 2021-04-01
gql_dio_link: Support specifying http statuses that are considered ok

jamonkko:
Depending of your graphql server the response statuses considered as valid graphql responses might differ, e.g. it responds always with 200.

### DIFF
--- a/links/gql_dio_link/lib/src/dio_link.dart
+++ b/links/gql_dio_link/lib/src/dio_link.dart
@@ -71,12 +71,15 @@ class DioLink extends Link {
   /// Dio client instance.
   final dio.Dio client;
 
+  final List<int> successStatuses;
+
   DioLink(
     this.endpoint, {
     @required this.client,
     this.defaultHeaders = const {},
     this.serializer = const RequestSerializer(),
     this.parser = const ResponseParser(),
+    this.successStatuses,
   }) : assert(client != null);
 
   @override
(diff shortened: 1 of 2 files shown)


agent3bood:
@jamonkko Thanks for contributing to this library.

The success status codes can be from 200 to 299, would it be better to chat that range only instead of adding new property to the constructor? For example are there any server that consider codes outside this range as success.

jamonkko:
The reasoning behind this would be that many graphQL servers return always 200. I know graphQL itself does not care about http layer and you could also return other than 200, but if I know that when server responds with graphql payload then the code is always 200 then I would like to verify it and otherwise interpret the response as non valid.

But this is also not very important feature for me :) I ended up adding this when debugging some problems, and thought to offer it also as PR. 
We could also just pass this change for now if it does not seem that relevant.

links/gql_dio_link/lib/src/dio_link.dart
```diff
@@ -91,7 +94,10 @@ class DioLink extends Link {
       },
     );
 
-    if (dioResponse.statusCode >= 300 ||
```
jamonkko, line 94: It might be that this check is always false (and not needed) if I understood correctly that Dio throws error (instead of returning a response) if status code is >= 300. 

Anyway, adding that optional check of valid status codes here would not change behaviour at all if you don't provide the optional status codes in constructor.

agent3bood:
@jamonkko Thanks for the feedback, I am not very familiar with how Did client works, I will look into how it handles errors.

agent3bood:
This is out of the scope of this package, modifying response status code can be done using dio interceptors.

```
  final client = dio.Dio()
    ..interceptors.add(DioCacheInterceptor(options: options))
    ..interceptors.add(dio.InterceptorsWrapper(
        onRequest:(options, handler){
          return handler.next(options); 
        },
        onResponse:(response,handler) {
          response.statusCode = 200; // << change statusCode
          return handler.next(response);
        },
        onError: (dio.DioError e, handler) {
          return  handler.next(e);
        }
    ));

```

PR #240, opened 2021-06-12
gql_websocket_link:  better (correct) assertion

samuelstroschein:
The assertion now checks if either one of the two values is null while the other is not which aligns with the given documentation. The previous assertion only checked if one of them is null. 

Took me quite some time to find the bug which was caused by a null URI in an isolate.

### DIFF
--- a/links/gql_websocket_link/lib/src/link.dart
+++ b/links/gql_websocket_link/lib/src/link.dart
@@ -120,7 +120,8 @@ class WebSocketLink extends Link {
     this.graphQLSocketMessageDecoder = _defaultGraphQLSocketMessageDecoder,
     this.initialPayload,
     this.inactivityTimeout,
-  }) : assert(uri == null || (channelGenerator == null)) {
+  }) : assert((uri == null && channelGenerator != null) ||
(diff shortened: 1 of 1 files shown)


links/gql_websocket_link/lib/src/link.dart
```diff
@@ -120,7 +120,8 @@ class WebSocketLink extends Link {
     this.graphQLSocketMessageDecoder = _defaultGraphQLSocketMessageDecoder,
     this.initialPayload,
     this.inactivityTimeout,
-  }) : assert(uri == null || (channelGenerator == null)) {
```
agent3bood, line 123: Please remove the extra parentheses.
aassert(uri == null || channelGenerator == null)`

samuelstroschein: What do you mean? This line is deleted/modified in this PR and does not include extra parentheses around channelGenerator anyways.

agent3bood: My mistake, I was confused by the change you made.

PR #249, opened 2021-07-09
Keep raw error data in GraphQLError

dmitryhryppa:
Hello!
Could we please keep raw error data in GrapQLError to access fields from the error response which was not represented by GraphQLError class itself? That would be very useful.

For example, I have some fields in the error response which I cannot access in current implementation and cannot change it on the backend side. Backends told me that this is totally valid to keep some extra fields in errors body, so I should workaround it somehow on the client side... So I hope this change could be merged :) 

Thank you!


dmitryhryppa:
Hello! 
Any chance to merge this?

agent3bood:
Hello @dmitryhryppa 

Sorry for the long delay, I am checking the code for links and I want to know why the following is not working for you?

The `request` method will return `Stream<Response>` type,  the `Response` contains `errors, data, and context`. The `data` is the raw http response data, can you find what you are looking for in this field?

dmitryhryppa:
@agent3bood thank you for your reply!

`Response` represents errors as the list of: `final List<GraphQLError>? errors;`, and `GraphQLError` class serializing only this fields:
```dart
  final String message;
  final List<ErrorLocation>? locations;
  final List<dynamic /* String | int */ >? path;
  final Map<String, dynamic>? extensions;
```
But in my case, the error itself contains more fields, the server responds to me with additional sub-errors and additional info inside the main error body. And these fields are not represented in `GraphQLError` class. This is designed on the server side and can not be changed from the client.
So to deal with it in a more generic way, I tried to store the unserialized raw error inside `GraphQLError`.
In that case, you can access additional fields of the error like `response.errors.first.raw['additionalField']`

Thank you!

Issue #269, opened 2021-09-24
Interfaces cannot implement other interfaces


ValentinVignal:
I'm generating my `schema.graphql` file with nestjs graphql: https://github.com/nestjs/graphql which supports interfaces implementing other interfaces:

```gql
interface Interface0 {
   field0: String
}

interface Interface1 implements Interface0 {
   field0: String
   field1: Float!
}

interface Interface2 implements Interface0 & Interface1 {
   field0: String
   field1: Float!
   field2: Float
}

type Type0 implements & Interface0 & Interface1 & Interface2 {
   field0: String
   field1: Float!
   field2: Float
}
```

But it looks like it is not supported by this package:

```
Error on line xx, column yy of package:my_project/schema.graphql: Unknown definition type 'implements'
   ╷
16 │ interface Interface2 implements Interface0 & Interface1 {
   │                      ^^^^^^^^^^
   ╵
```

(Right now, I'm using the version `0.9.4`)
Is there something I should do to make it work, or is it planned to support it in the future?

smkhalsa:
First of all, I moved this to the `gql` repo since it relates to the gql/language code, not `ferry` directly.

The feature of interfaces implementing other interfaces hasn't made it into the official GraphQL spec yet.

They are [included](https://spec.graphql.org/draft/#sec-Interfaces.Interfaces-Implementing-Interfaces) in the current working draft though, so we can reasonably expect them in the next official spec release. As such, I think it would be reasonable to just go ahead and implement this feature now.

I don't have the time to work on this, but I can provide guidance if you'd like to make a PR, or maybe one of the other maintainers of `gql` wants to take it on.

ValentinVignal:
Awesome ,I'm very glad to hear that. Thank you for this quick response. I don't have the time to work on it right now though. I might give it a try in a while if no one started in the meantime.

tlvenn:
Would be great to add support for this, it is now part of the official spec since October 2021.

Issue #267, opened 2021-09-04
Support for VARIABLE_DEFINITION directive location


alexander-shustanov:
When trying to create document for schema with 
``` graphql
directive @dtoFor(type: String!) on VARIABLE_DEFINITION
```

an exception is raised:

```
Error on line 7, column 37: Unknown directive location 'VARIABLE_DEFINITION'
  ╷
7 │ directive @dtoFor(type: String!) on VARIABLE_DEFINITION
  │                                                
  ╵
```

VARIABLE_DEFINITION is a valid directive location actually.

PR #279, opened 2021-10-25
Support VARIABLE_DEFINITION and interfaces implement other interfaces

juancastillo0:
Attempts to fix [267](https://github.com/gql-dart/gql/issues/267) and [269](https://github.com/gql-dart/gql/issues/269).

Hi, this adds support for the "VARIABLE_DEFINITION" `DirectiveLocation.variableDefinition` and adds a test which parses directives definitions with all the `DirectiveLocation`s.

Also tries to support interfaces that implement other interfaces, with a test taken from [269](https://github.com/gql-dart/gql/issues/269) and using `printNode` to verify that multiple interfaces are printed on an interface definition. Not sure if that test should go in that file, specially the `printNode` part.

I added untested code in gql_code_build with the new additions, however I'm not sure if there's something else that should be done.

Thanks, if there's anything more I can do, please let me know


tlvenn:
@agent3bood any chance you can unblock this ? Thanks in advance

agent3bood (approved):

PR #286, opened 2021-11-19
link/websocket: Add connection state callback

julianscheel:
This is an approach to make connection tracking available to users of the `WebSocketLink`.
Instead of offering a callback one could also add a getter to make the `_connectionStateController`'s `stream` accessible, so that a user could directly `listen` to it.

I'd be happy about any comments.

### DIFF
--- a/links/gql_websocket_link/CHANGELOG.md
+++ b/links/gql_websocket_link/CHANGELOG.md
@@ -1,4 +1,4 @@
-0.3.2
+## 0.3.2
 
 - Upgrade `gql_exec`.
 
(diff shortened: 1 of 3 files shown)


agent3bood:
Thanks for contributing to the project.
> add a getter to make the _connectionStateController's stream accessible

This would be cleaner way to expose the connection state to the user.

julianscheel:
@agent3bood Sorry for the long delay. I got finally back to this.
The callback is replaced with a getter for the `_connectionStateController`'s `stream` now. Additionally I replaced the `const int` state value with a dedicated `enum` to have a more concise api.
Would you mind taking another look?

agent3bood (approved):

agent3bood:
Thanks for updating this 
Would you upgrade the version on pubspec to 0.3.3 and add an entry to CHANGELOG.md

julianscheel:
@agent3bood Added a 0.3.3 bump commit

julianscheel:
Nice, thanks! Will you release 0.3.3 to pub.dev?

PR #298, opened 2022-02-18
support analyzer ^3.0.0

knaeckeKami:


agent3bood:
Hi @knaeckeKami 

Thanks for the PR, I am looking at [Analyzer 3.0.0](https://pub.dev/packages/analyzer/changelog#300), and it did remove many deprecated functions. So I am not sure supporting 2 major versions will work `analyzer: '>=2.3.0 <4.0.0'`.

agent3bood:
> I have to check if I can get the exact required version of graphviz installed on my mac machine, it seems brew only has newer versions unfortunately
Never mind  the svg check, it has been failing for some time now!

knaeckeKami:
Ok, well I think this is ready then!

agent3bood:
Please change the package version to `version: 0.5.0` and add an entry to CHANGELOG.

agent3bood (approved):

budde377:
@agent3bood thank you for this! When are you expecting to release it to pub?

agent3bood:
> @agent3bood thank you for this! When are you expecting to release it to pub?

@klavs Can you upgrade from alpha release?

knaeckeKami:
if you are using a ferry, you probably cannot upgrade yet. I think there is an issue with ferry_generator, (the serializer for FetchPolicy might not be imported correctly ).

I probably cannot look into it in the next few days, but I think the fix should be trivial.

agent3bood:
I will try to solve this tomorrow as we need it for a project, just to same me some time are there other packages need upgrading?

klavs:
@agent3bood, please check my message on discord.

Issue #242, opened 2021-06-15
Unable to determine the response type of a union when using obfuscated code


DustinJSilk:
When returning a possible union from a GraphQL endpoint, we need to check the type of the object that is returned to correctly act on it.

For example, with the following query:

```GraphQL
query getData {
  getData {
    ... on Success {
      data
    }

    ... on Error {
      message
      code
    }
  }
}
```

We can check the type using the follow code in a request response:

```dart
if (data.getData is GgetDataData_getData__asSuccess) {
  ...
} else if  (data.getData is GgetDataData_getData__asError) {
  ...
}
```

This works correctly in both the editor with intellisense and when running the app in a simulator/emulator.

The problem is that it doesn't work when building a `--release` and a `--obfuscate --split-debug-info=./debug_split` app running on a physical device. This means we have to disable code obfuscation to support unions.

I've create a repository with the steps to reproduce the issue here: [https://github.com/DustinJSilk/gql_unions_test](https://github.com/DustinJSilk/gql_unions_test)

Apologies if this should be in the `gql-dart/gql` repository, I'll add a ticket there too and remove one if necessary.

smkhalsa:
The issue has been transfered.

DustinJSilk:
Okay!, 
I'm going to try and do some digging around for techniques to get around this issue. This seems to work with other libraries like flutter_bloc so I'm not exactly sure what the difference is here with these type checks.

If I find a solution I'll submit a PR

bgparkerdev:
I think this should be looked at as anyone who uses flutter web will run into this issue. 

It seems like the issue might be in [here](https://github.com/gql-dart/gql/blob/b761664f75834e4487a8c1d6f7bfd9a1c84adf5e/codegen/gql_code_builder/lib/src/serializers/inline_fragment_serializer.dart) . It looks like `_typeForTypename` , is doing a `.toString()` on the type, which cannot be relied on. See [source](https://github.com/flutter/flutter/issues/66236#issuecomment-698228855).

bgparkerdev:
@smkhalsa I tried to do something in https://github.com/gql-dart/gql/pull/312 . I'm still very new to dart code gen stuff, and was having troubles figuring out how to run all the tests. So any help would be greatly appreciated.

PR #312, opened 2022-04-28
Union types in web release mode

bgparkerdev:
Release builds on web do code minification, this causes serialization issues when inline fragments are used. 

The InlineFragmentSerializer uses `toString()` on a type to compare to a graphql `__typename`. Unfortunately in minifed versions the `toString()` method on a dart Type returns a mangled class name and not the expected class name.


bgparkerdev:
Resolves #242 @smkhalsa

bgparkerdev:
Fixed the failing e2e test related to my changes.

bgparkerdev:
Used multipack and build_runner to generate the .gql file that needed to be updated. I noticed that running `build_runner` produced a lot of changes in the `.gql` files, perhaps it has not been ran in this repo in a while?

bgparkerdev:
Can someone kick off the tests for me again? Thanks.

PR #317, opened 2022-06-28
escape names used by built_value

knaeckeKami:
toJson / fromJson and serializer is used by Built classes, build and replace by Builder classes.
Using them as field names in a graphql schema breaks code generation currently.


ebsangam:
@knaeckeKami Please check if end_to_end_test passes with this commit.

PR #319, opened 2022-07-18
rxdart updated to latest version

kenspy:
fix dependency version mismatch with other packages.


knaeckeKami:
could you leave the version range open? 

Since both rxdart 0.26 and 0.27 seem to work, you could set it to

```yaml
rxdart:  '>=0.26.0 <0.28.0'
```

to decrease the chance of dependency conflicts with other libraries

PR #323, opened 2022-07-24
Analyzer update & inList check removal.

2shrestha22:
- Updated analyzer dependency to `>=4.2.0 <5.0.0`
- Removed inList check, https://github.com/google/built_value.dart/issues/1011#issuecomment-804843573, https://github.com/gql-dart/gql/pull/202

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -83,7 +83,7 @@ jobs:
       - name: Activate multipack
         run: |
           echo "$HOME/.pub-cache/bin" >> $GITHUB_PATH
-          pub global activate multipack
+          dart pub global activate multipack
       - name: Check pubspec
         run: |
           multipack --only $PACKAGE pubspec clean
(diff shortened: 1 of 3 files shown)


2shrestha22:
@knaeckeKami How to solve these? Seems related to incompatible SDK version.

2shrestha22:
CI uses discontinued dart image. Using newer docker image will solve pass these tests. Currently using dart image which is discontinued -https://hub.docker.com/r/google/dart/ , one with latest update which we should use - https://hub.docker.com/_/dart.

2shrestha22:
@knaeckeKami I have updated the dart docker image. Would you please re-run the tests?

codegen/gql_build/lib/src/utils/config.dart
```diff
@@ -10,7 +10,7 @@ Map<String, Reference> typeOverrideMap(dynamic typeOverrideConfig) {
           entry.key as String,
           Reference(
             entry.value["name"] as String,
-            entry.value["import"] as String,
+            entry.value["import"] as String?,
```
ebsangam, line 13: @knaeckeKami Please have a look here, since `entry.value["import"]` was null and I was facing issue with code gen, I made it nullable. I am not sure what it does and why it was not nullable type before.

Issue #295, opened 2022-02-03
Add option to `gql_dio_link` to support GET requests


ueman:
Currently, the [`gql_dio_link`](https://pub.dev/packages/gql_dio_link) does not support doing request via GET. I would like see to support for that.

PR #324, opened 2022-07-28
[gql_dio_link] Add ability to use get for queries

ueman:
This adds the ability to use get requests for queries, similar as the implementation in the `gql_http_link`.

Let me know if there's anything missing and I'll follow up on it.

Closes https://github.com/gql-dart/gql/issues/295


ueman:
Are the CI failures related to my changes? As far as I can tell, they aren't, but the logs aren't really insightful.

knaeckeKami:
No, I don't think so. The check_svg pull request fails because the SVG changed. 

I think I will delete this check as it is very flaky and fails when the tool version was updated or run from a different OS. 
The end to end test seems to be broken atm (unrelated to your PR). Please let me investigate this first.

knaeckeKami:
Hey, publishing gql_dio_link with this PR merged failed, see https://github.com/gql-dart/gql/runs/7654194681?check_suite_focus=true

Can you take a look please?
I think it's just a matter of moving dependencies from dev_dependencies to dependencies: in pubspec.yaml

ueman:
> Hey, publishing gql_dio_link with this PR merged failed, see https://github.com/gql-dart/gql/runs/7654194681?check_suite_focus=true
> 
> Can you take a look please? I think it's just a matter of moving dependencies from dev_dependencies to dependencies: in pubspec.yaml

Yes, I'll follow up in another PR and ping you when I'm ready.

PR #329, opened 2022-08-01
chore: use latest dart docker image

2shrestha22:
CI was using discontinued dart image. Using newer docker image will solve pass these tests. Currently using dart image which is discontinued -https://hub.docker.com/r/google/dart/ , one with latest update which we should use - https://hub.docker.com/_/dart.


2shrestha22:
Moved https://github.com/gql-dart/gql/pull/328

PR #330, opened 2022-08-03
Fix dependencies for gql_dio_link

ueman:
This fixes the error mentioned in https://github.com/gql-dart/gql/pull/324#issuecomment-1204051957

I also suggest adding `depend_on_referenced_packages` to the lints.

cc @knaeckeKami


knaeckeKami:
Thanks! I agree, the lints should be improved.

PR #331, opened 2022-08-03
Add original stacktrace to exceptions with original exceptions

### LINKED ISSUE #296: `LinkException`s should contain stacktraces
Currently, all of the `LinkException`s do not contain the stacktraces of the original exceptions. This can make it hard to figure out why or where an exception was thrown.

ueman:
Closes #296

This adds the original stacktrace to exceptions. It improves the debugging experience, and it's especially helpful if those exceptions are reported via crash reporting tools.
Technically, this is a breaking change.

This would be really nice to have for https://github.com/ueman/sentry-dart-tools/tree/main/sentry_link


knaeckeKami:
LGTM

PR #333, opened 2022-08-04
Fix Dio Link with get requests and add more tests

ueman:
I've noticed that the `DioLink` wasn't working in all cases. It works now much more similar than the `HttpLink` with get requests, which was working for all cases I know of. While fixing this, I've made the test more extensive.


ueman:
@knaeckeKami Thanks for the swift reviews/merges. It's a pleasure to contribute so far

Issue #339, opened 2022-08-06
Graphviz dot produces indeterminsitic result, breaking the CI SVG check


knaeckeKami:
```bash
repeat 20 multipack info | dot -Tsvg -x | md5
```
outputs
```
2db5aa6b11b6b4f52e8283f2435d5212
d6156c52000e690a7c3e72e52553b9db
2db5aa6b11b6b4f52e8283f2435d5212
2db5aa6b11b6b4f52e8283f2435d5212
1979929664b2fbc1bce4975000f37ac3
5201338dd5c626e72645bbcedf9ea4bf
5201338dd5c626e72645bbcedf9ea4bf
2db5aa6b11b6b4f52e8283f2435d5212
5201338dd5c626e72645bbcedf9ea4bf
2db5aa6b11b6b4f52e8283f2435d5212
2db5aa6b11b6b4f52e8283f2435d5212
2db5aa6b11b6b4f52e8283f2435d5212
5201338dd5c626e72645bbcedf9ea4bf
5201338dd5c626e72645bbcedf9ea4bf
5201338dd5c626e72645bbcedf9ea4bf
5201338dd5c626e72645bbcedf9ea4bf
5201338dd5c626e72645bbcedf9ea4bf
2db5aa6b11b6b4f52e8283f2435d5212
5201338dd5c626e72645bbcedf9ea4bf
1979929664b2fbc1bce4975000f37ac3
```

on my machine. 
It should output the same hash every time if the algorithm was deterministic.
The output of `multipack info` is deterministic, so it must be `dot` which is the culprit here. 
The actual difference in the SVG is tiny, it seems to come down to some rounding differences.

I believe this is a known issue of `dot` discussed here: https://gitlab.com/graphviz/graphviz/-/issues/1767

Because of this, the check SVG step in the CI does not really make sense in my opinion, 
since it would often fail even though the  SVG image is up to date.

PR #352, opened 2022-08-26
Disable lint on generated files

2shrestha22:
Related issue: https://github.com/gql-dart/ferry/issues/388
https://github.com/dart-lang/linter/issues/320

To avoid lints on generated files we have to exclude generated files from analyzer but it also disables static analysis.
https://github.com/dart-lang/sdk/issues/46957

### DIFF
--- a/codegen/end_to_end_test/lib/aliases/__generated__/aliased_hero.ast.gql.dart
+++ b/codegen/end_to_end_test/lib/aliases/__generated__/aliased_hero.ast.gql.dart
@@ -1,4 +1,5 @@
 // GENERATED CODE - DO NOT MODIFY BY HAND
+// ignore_for_file: type=lint
 
 // ignore_for_file: no_leading_underscores_for_library_prefixes
 import 'package:gql/ast.dart' as _i1;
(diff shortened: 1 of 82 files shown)


knaeckeKami:
please re-run codegen and add to your PR locally so that the examples have the right generated code checked in

otherwise LGTM

2shrestha22:
@knaeckeKami I think multipack is not working properly. I did `multipack exec dart pub run build_runner build --delete-conflicting-outputs` put codes are not generated with new changes.

knaeckeKami:
You need to multpack pubspec override before, other it will not use the local dependencies but the ones from pub. 
But remember to call multpack clean before committing

2shrestha22:
@knaeckeKami I think new version of `graphviz` is available. But why should we diff svg instead of graph itself?

knaeckeKami:
the point of the svg check is, that the generated svg file is always at the latest state in the README

Checking the .dot does not guarantee this. We would have the generate the SVG from the dot file and then push a new commit with the latest svg on every MR.

But for now because of https://github.com/gql-dart/gql/issues/339, this does not work anyway. I'll probably deactivate the check if I don't find a robust solution.

2shrestha22:
thank you

PR #359, opened 2022-10-21
chore: updated analyzer

amondnet:
Set version constraints for analyzer to `>=4.6.0 <6.0.0`


knaeckeKami:
Thanks. Some of the checks failed because because the generated files are formatted differently. I can reproduce this by using the latest version of dart vs the stable version, I suspect it's because of an update to dart format.

The changes are purely in the formatting though, so they don't really matter. 
LGTM.

Issue #361, opened 2022-10-25
Unify code duplication of `gql_http_link` and `gql_dio_link`


ueman:
Both, `gql_http_link` and `gql_dio_link`, share a fair bit of implementation. It would be nice if it could be somehow shared between those two.

The following code snippets are mostly the same:

https://github.com/gql-dart/gql/blob/c13e2fe2e6ca234ff5bc17263e3536ed2c907474/links/gql_http_link/lib/src/link.dart#L16-L52

https://github.com/gql-dart/gql/blob/c13e2fe2e6ca234ff5bc17263e3536ed2c907474/links/gql_dio_link/lib/src/dio_link.dart#L11-L43

Then there's also https://github.com/gql-dart/gql/issues/360

I'm guessing there's also some more, if one searches for it.

knaeckeKami:
Yes, it might be possible to extract a lot of the login in same gql_http package and let both http_link and dio_link depend on it.

PR #364, opened 2022-11-03
Add `BaseHttpLinkHeaders` and `BaseHttpLinkResponseContext` to gql_exec

ueman:
This is a first step towards fixing #361. After this is merged and release, I'll change the contexts in `gql_http_link` and `gql_dio_link` to be subclasses of the classes introduced in this PR. Therefore, it will be a non-breaking change. 

If you have any other ideas on how to introduce this in a non-breaking change, feel free to suggest them.


knaeckeKami:
I like the idea of keeping this a non-breaking change by introducing base types.

PR #369, opened 2022-11-12
fix: use typedefs instead of subclasses for base contexts for http bases links

knaeckeKami:
@ueman fyi

### DIFF
--- a/links/gql_exec/lib/src/http.dart
+++ b/links/gql_exec/lib/src/http.dart
@@ -32,7 +32,8 @@ class HttpLinkResponseContext extends ContextEntry {
   /// HTTP headers of the response. if you need duplicated headers, use [rawHeaders]
   Map<String, String>? get headers =>
       _headers ??
-      _rawHeaders?.map((key, valueList) => MapEntry(key, valueList.isEmpty ? "" : valueList.last));
+      _rawHeaders?.map((key, valueList) =>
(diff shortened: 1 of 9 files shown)


links/gql_exec/lib/src/http.dart
```diff
@@ -21,16 +21,35 @@ class BaseHttpLinkHeaders extends ContextEntry {
...
+
+  /// HTTP headers
+  Map<String, List<String>>? get rawHeaders =>
+      _rawHeaders ?? _headers?.map((key, value) => MapEntry(key, [value]));
+
+  const HttpLinkResponseContext({
     required this.statusCode,
-  });
+    Map<String, String>? headers,
+    Map<String, List<String>>? rawHeaders,
```
ueman, line 45: Why do we need both of these header maps?

knaeckeKami: http headers can be repeated. [package:http does not support this](https://github.com/dart-lang/http/issues/24), but dio does. See **https://www.rfc-editor.org/rfc/rfc2616.html** 

So the alternatives are:

- only adding the  Map<String, List<String>> headers, but this would be a breaking change, and also would make the API more awkward for use cases that don't need repeated headers
- not support repeated headers at all
- add both

ueman: Huh TIL

ueman (approved):

Issue #372, opened 2022-11-23
bug [codegen]: Nesting InlineFragmentNode on an extended-Type in a FragmentDefinition based on an Interface Type definition causes unbound recursion/Stack_Overflow


JELaVallee:
We ran into this issue in our Flutter project where we've been using Fragments to very effectively share GraphQL sub-graph models between Widgets. 

We've recently implemented a feature with a Query Field that returns an Interface instead of the concrete-Type of the parent Node. Suddenly, we were getting `Stack Overflow` errors thrown when running codegen...

Our use case is similar to this scenario:

```graphql
query HeroWithFragments($episode: Episode) {
  hero(episode: $episode) {
    ...characterFieldsFragment
  }
}

fragment characterFieldsFragment on Character {
  id
  name
  ... on Human {
    ...humanFieldsFragment
  }
}

fragment humanFieldsFragment on Human {
  homePlanet
}
```

When I run codegen on this (in the `codegen/end_to_end_test` sub-proj w/ pub overrides enabled _or_ with just our normal `build_runner` setup), the build run fails with:

<details>
<summary>ERR_OUT: `dart pub run build_runner build --delete-conflicting-outputs -v`</summary>

```
[FINE] built_value_generator:built_value on lib/fragments/__generated__/hero_with_fragments.var.gql.dart:Running BuiltValueGenerator
lib/…/__generated__/hero_with_fragments.var.gql.dart:1
[FINE] built_value_generator:built_value on lib/variables/__generated__/human_with_args.var.gql.dart:Running BuiltValueGenerator
lib/…/__generated__/human_with_args.var.gql.dart:1
[FINE] built_value_generator:built_value on lib/variables/__generated__/create_review.var.gql.dart:Running BuiltValueGenerator
lib/…/__generated__/create_review.var.gql.dart:1
[FINE] built_value_generator:built_value on lib/aliases/__generated__/aliased_hero.var.gql.dart:Running BuiltValueGenerator
lib/…/__generated__/aliased_hero.var.gql.dart:1
[FINE] built_value_generator:built_value on lib/interfaces/__generated__/hero_for_episode.var.gql.dart:Running BuiltValueGenerator
lib/…/__generated__/hero_for_episode.var.gql.dart:1
[FINE] built_value_generator:built_value on lib/graphql/__generated__/serializers.gql.dart:Running BuiltValueGenerator
lib/…/__generated__/serializers.gql.dart:1
[INFO] Build:Running build completed, took 3.5s

[...]

Issue #228, opened 2021-05-20
Fragment spreads on a union type cause an error


tberman:
Due to the way fragment spreads are handled (by walking them and including them in the parent selection set) when you have a fragment spread on a union you get an error, for example:

```
query SearchQuery {
  search {
    ...DroidFragment
  }
}

fragment DroidFragment on Droid {
  name
}
```

causes the error

```
Exception: SearchResult is not an ObjectTypeDefinitionNode or InterfaceTypeDefinitionNode
```

It appears that fragment spreads should be handled more like inline fragments and have built subclasses, but I am likely missing something.

Happy to attempt to put together a PR if there is consensus on what the right behavior would be!

agent3bood:
Hi

Would you please share the schema, or provide a complete example to try to reproduce it.
I have used Fragments and it is just the way you did it.

tberman:
The example provided will fail against the swapi schema used in the `end_to_end` examples in this repo. I can push up a branch that shows the failure. It's specifically from fragment spreads on unions not wrapped inside an inline fragment. You can see in some other examples in `end_to_end` where the spread is being wrapped in an inline fragment to work around this.

tberman:
Pushed up a [branch](https://github.com/gql-dart/gql/compare/master...tberman:union_test_case) that shows the issue.

If you run build, you will get the error I mentioned.

I did some quick digging and if you look at how https://github.com/gql-dart/gql/blob/master/codegen/gql_code_builder/lib/src/operation/data.dart#L124 handles FragmentSpreads it hoists them onto the parent, causing this (correct) error for unions. Instead, it seems like it should be handled as the function handles InlineFragmentNodes: https://github.com/gql-dart/gql/blob/master/codegen/gql_code_builder/lib/src/operation/data.dart#L160

tberman:
@agent3bood I am happy to spend some time poking at this if we get to a place where there is some general agreement on the right solution!

agent3bood:
Hello again 

First I want to point to some resources, where in both documentations the way to select a union is to use inline fragment, which is working fine. 

https://graphql.org/learn/queries/#inline-fragments
https://www.apollographql.com/docs/android/essentials/inline-fragments/




*working as per standards*
```
query Search {
    search {
        ... on Human {
            name
        }
        ... on Droid {
            ...DroidFragment
        }
    }
}

fragment DroidFragment on Droid {
    name
}
```




I am not sure wether the suggested way is a standard or a `shorthand` developed by the community.

[...]

PR #373, opened 2022-11-29
bugfix [codegen]: Proper field selector recursion on InlineFragments for Interface-extended Types and Typed FragmentSpreads  

JELaVallee:
## Summary

This is a fix for #372 where this scenerio for creating fragments would fail during codegen with a stack-overflow:

```graphql
query HeroWithInterfaceSubTypedFragments($episode: Episode!) {
  hero(episode: $episode) {
    ...heroFieldsFragment
  }
}

fragment heroFieldsFragment on Character {
  id
  name

  ...on Human {
    ...humanFieldsFragment
  }

  ...on Droid {
    ...droidFieldsFragment
  }
}

fragment humanFieldsFragment on Human {
  homePlanet
  birthday
  friends {
    ...on Droid {
      ...droidFieldsFragment
    }
  }
}

fragment droidFieldsFragment on Droid {
  primaryFunction
}
```

The issue was traced to the data builder phase of the build process in this [block of code](https://github.com/gql-dart/gql/blob/master/codegen/gql_code_builder/lib/src/operation/data.dart#L299-L304):

```dart
            if (retainFragmentSpreads) selection,
            ..._expandFragmentSpreads(
              fragmentMap[selection.name.value]!.selections,
              fragmentMap,
              false,
            )
```

This will effectively cause an unbound recursion of the node-tree from this point forward, appending on another `__base` definition to the builder-tree for the currently evaluated node before doing another traversal.  Basically causing nested-`__base` definitions for the builder... all the way down. 😉 

[The fix we've come up with](https://github.com/gql-dart/gql/blob/jelavallee/fix-issue372-inline-frag-nesting-failure/codegen/gql_code_builder/lib/src/operation/data.dart#L299-L313) adds filtering of the subfields prior to the next recursion step and doesn't pass the `InlineFragment` of current Node, just the "branches and leaves". 

```dart
          final fragmentSelections =
              fragmentMap[selection.name.value]!.selections;

          return [
            if (retainFragmentSpreads) selection,
            ..._expandFragmentSpreads(
              [
                ...fragmentSelections.whereType<FieldNode>(),
                ...fragmentSelections.whereType<FragmentSpreadNode>(),
              ],
              fragmentMap,
              false,
            )
          ];
```
This allows for proper traversal/termination for the `FragmentSpread` and `FieldNode` nodes without falsely re-evaluating the `InlineFragment`'s base-Interface fields. The generated code also has the correct sub-typing of Data class definitions and serializers.

### Notes: 

1. This does not fix [a similar issue with Fragments typed against Unioned Types](https://github.com/gql-dart/gql/issues/228).  Not sure on this one... Union is like an Interface without a base-definition. 
2. This does not fix an issue where the data builder fails to parse the extended-Type of a `FragmentDefinition` when coming from and scope that is of a base-Interface. (for example, removing the `...on Human` inline in the `fragment heroFieldsFragment on Character` definition in the above scenario, will cause the builder to treat `humanFieldsFragment` as being of `Character` type scope. That causes a `Bad State: not found` exception when attempting to resolve a sub-Typed field. (will be investigating this in a future issue/PR!)

## Testing

Currently, I've added the above scenario to the `codegen/end_to_end` testing harness and confirmed that the build_runner completes generating the corresponding gql-dart artifacts without failure.

This scenario also includes doubled-nesting of typed Fragments to ensure that the solution works at multiple depths.

WIP: Adding some test cases in the [codegen/end_to_end_test/test/operation/fragments_test.dart](https://github.com/gql-dart/gql/blob/jelavallee/fix-issue372-inline-frag-nesting-failure/codegen/end_to_end_test/test/operation/fragments_test.dart#L1) tests.


knaeckeKami:
Hi, thanks for this PR!

On the first glance, this looks valid to me. Will to a deeper dive into it later.

> This does not fix an issue where the data builder fails to parse the extended-Type of a FragmentDefinition when coming from and scope that is of a base-Interface. (for example, removing the ...on Human inline in the fragment heroFieldsFragment on Character definition in the above scenario, will cause the builder to treat humanFieldsFragment as being of Character type scope. That causes a Bad State: not found exception when attempting to resolve a sub-Typed field. (will be investigating this in a future issue/PR!)

Inline Fragment spreads without a type condition are not supported at the moment. 
There are a few TODOs `/// TODO: Handle inline fragments without a type condition`. 
This might be part of the issue.

JELaVallee:
> Inline Fragment spreads without a type condition are not supported at the moment. There are a few TODOs `/// TODO: Handle inline fragments without a type condition`. This might be part of the issue.

Ah good to know! I did some debugging into while I was chasing this current issue, and it seemed like the `FragmentDefinitions` would have to be parsed first and available during `OperationDefinition` tree-walk in order to "inject" proper `TypedNode`s in the tree based on the `FragmentDefinition`.  Didn't dig much past seeing that. 🕦

codegen/end_to_end_test/test/operation/fragments_inline_nesting_test.dart
```diff
@@ -0,0 +1,41 @@
...
+
+      final gqlHuman = GheroFieldsFragmentData__asHuman.fromJson(jsonHuman);
+
+      expect(gqlHuman?.id, equals("luke"));
+      expect(gqlHuman?.name, equals("Luke"));
+      expect(gqlHuman?.homePlanet, equals("Tattoine"));
+      expect(gqlHuman?.birthday, equals(DateTime.parse("1976-05-04T08:00:00")));
+      expect(gqlHuman?.friends?.first?.G__typename, equals("Droid"));
+      // FAILS: "The getter 'primaryFunction' isn't defined for the type 'GheroFieldsFragmentData__asHuman_friends'."
+      // expect(gqlHuman?.friends?.first?.primaryFunction, equals("Astromech"));
```
JELaVallee, line 38: Will be looking into it a bit deeper. 

Debugging into this, there are `Droid` specific fields in the `gqlHuman.friends` collection's items, BUT, those fields aren't accessible because the return type of the `friends` collection is for a narrower Type. (that is, it should be `_GheroFieldsFragmentData_asHuman_friends__asDroid`)

knaeckeKami: I think this is expected behavour. Since it's possible that there are friends who are not droids but human, the return type has to be the base class with that has only the shared fields. To access the special fields, a type check or a cast is needed like

```dart
   final first = gqlHuman?.friends?.first as GheroFieldsFragmentData__asHuman_friends__asDroid;
   expect(a.primaryFunction, equals("Astromech"));
```

JELaVallee: Yep... I wasn't adding the type-filter... Just updated that!

Also extended this to what my original intent which was to demonstrate sub-type Inlines in the sub-Fragment.

codegen/end_to_end_test/lib/fragments/hero_with_interface_subtyped_fragments.graphql
```diff
@@ -0,0 +1,41 @@
...
+fragment humanFieldsFragment on Human {
+  homePlanet
+  birthday
+  friends {
+    ...on Droid {
+      id
+      name
+      ...droidFieldsFragment
+    }
+    #### NOTE: This cannot self-reference ...humanFieldsFragment hench the expanded fields
```
JELaVallee, line 29: Pretty sure we're not supporting recursive fragments... But wanted to test this out. ;-)

codegen/end_to_end_test/test/operation/fragments_inline_nesting_test.dart
```diff
@@ -0,0 +1,63 @@
...
+      expect(gqlHuman_droidFriends?.first.id, equals("r2d2"));
+      expect(gqlHuman_droidFriends?.first.name, equals("R2-D2"));
+      expect(gqlHuman_droidFriends?.first.primaryFunction, equals("Astromech"));
+
+      final gqlHuman_humanFriends = gqlHuman?.friends
+          ?.whereType<GheroFieldsFragmentData__asHuman_friends__asHuman>();
+
+      expect(gqlHuman_humanFriends?.first.id, equals("han"));
+      expect(gqlHuman_humanFriends?.first.name, equals("Han Solo"));
+      expect(gqlHuman_humanFriends?.first.homePlanet, equals("Nerf"));
```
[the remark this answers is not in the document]
knaeckeKami: The code and tests look good to me now! 
I will properly review it this weekend and run it against my production app to check for possible regressions and then, if nothing comes up, merge it :)

PR #374, opened 2022-12-20
Feature/add interceptor

phucgaoxam:


knaeckeKami:
Thanks for your PR!

I would like to keep the links as minimal as possible. 
You can add HTTP headers by creating a link that add HttpLinkHeaders to the context of the Request object.

See for example this: https://github.com/gql-dart/gql/blob/master/examples/gql_example_http_auth_link/lib/http_auth_link.dart#L42

phucgaoxam:
sorry I wanted to create a PR in my folk, let me close this PR

PR #388, opened 2023-02-19
[gql_websocket_link] enhancement for connection errors

dehypnosis:
Hi,

At First, Currently Subscription requests are losing message streams when having WebSocketChannelException.
And this problem is cannot be recovered automatically even if autoReconnect is set true, so the websocket connection recovered later.
It means when client have a connection problem, all message stream subscriptions for 'Subscription Operation' should be manually recreated.
Fixed this issue by not closing message subscription on WebSocketChannelException.

Secondly, Here added `StreamController<ConnectionError> _connectionErrorController` and public accessor `connectionErrorStream`.
A practical use case is that, when web socket connection failed due to server side decision or unexpected error, client can be notified about detail of error to take some action.
For example from my case, I need to refresh my client-side id token when it has been expired. I am using Firebase Authentication, and I can make Firebase detect token expiration on App startup. But when app became active from background after more than one hour, I should be detect token is expired or revoked by server..

I made a PR to deal with above case, or any other server-side connection errors.
Please take a consideration of this PR and thank you for reviewing it.

### DIFF
--- a/codegen/gql_build/pubspec.yaml
+++ b/codegen/gql_build/pubspec.yaml
@@ -20,3 +20,7 @@ dependencies:
 dev_dependencies: 
   build_test: ^2.0.0
   gql_pedantic: ^1.0.2
+topics: 
+  - graphql
+  - gql
+  - codegen
--- a/codegen/gql_code_builder/pubspec.yaml
+++ b/codegen/gql_code_builder/pubspec.yaml
@@ -17,3 +17,7 @@ dev_dependencies:
   build_runner: ^2.1.0
   gql_pedantic: ^1.0.2
   test: ^1.16.8
+topics: 
+  - graphql
+  - gql
+  - codegen
--- a/gql/pubspec.yaml
+++ b/gql/pubspec.yaml
@@ -12,3 +12,6 @@ dependencies:
 dev_dependencies: 
   gql_pedantic: ^1.0.2
   test: ^1.16.2
+topics: 
+  - graphql
+  - gql
--- a/links/gql_dedupe_link/pubspec.yaml
+++ b/links/gql_dedupe_link/pubspec.yaml
@@ -14,3 +14,7 @@ dev_dependencies:
   gql_pedantic: ^1.0.2
   mockito: ^5.0.0-nullsafety.7
   test: ^1.16.2
+topics: 
+  - graphql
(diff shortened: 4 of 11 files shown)


knaeckeKami:
Thanks for your PR. 
It looks like the streamcontroller created is never disposed,  even it the link is disposed. could you fix that?

agent3bood:
Thanks for taking time to describe the issue and providing a PR.

We are already forwarding the errors to the same stream here, and you can listen to it as well.
With your addition does that bring new functionality to the library?
```
onError: (Object error) {
        _messagesController.addError(error);
 });
```

links/gql_websocket_link/lib/src/link.dart
```diff
@@ -405,6 +417,7 @@ class WebSocketLink extends Link {
     await _channel?.sink.close(websocket_status.normalClosure);
     _connectionStateController.add(ConnectionState.closed);
     await _connectionStateController.close();
+    await _connectionErrorController.close();
```
dehypnosis, line 420: Added disposing statement here.

links/gql_websocket_link/lib/src/link.dart
```diff
@@ -245,6 +251,8 @@ class WebSocketLink extends Link {
             ).catchError(_messagesController.addError);
           });
           _reConnectRequests.clear();
+        } else if (parsedMessage is ConnectionError) {
+          _connectionErrorController.add(parsedMessage);
```
dehypnosis, line 255: @agent3bood Thank you for the review.

As you mentioned, `onError` handler is already forwarding stream exceptions.
But `ConnectionError` seems not considered as an exception. And channel's `listen` handler currently handles `ConnectionAck` but not handle `ConnectionError`. So I added public connection error stream for the use-case handling ConnectionError manually.
https://github.com/gql-dart/gql/blob/master/links/gql_websocket_link/lib/src/link.dart#L237

Additionally, as I described in "First issue", when `_messagesController` got an 'error', all the subscription operations streams are closed now. So they cannot be recovered even after reconnection success.
So I added 277-280 Lines.

dehypnosis:
And due to CI check failure (formatting error), I pushed reformatted codes.
But I don't know how to re-run CI checks... Could you help me guys?

Issue #394, opened 2023-04-12
`Visitor` and all sub class could have a `const` constructor to always subclasses to be const


ValentinVignal:
It would be nice to add `const` constructors when applicable to `Visitor` and its subclasses so classes that extend them can have a const constructor

PR #395, opened 2023-04-12
revamp: Add `const` constructors to `Visitor`s

ValentinVignal:
Fixes https://github.com/gql-dart/gql/issues/394

### DIFF
--- a/gql/pubspec.yaml
+++ b/gql/pubspec.yaml
@@ -1,5 +1,5 @@
 name: gql
-version: 0.15.0
+version: 0.14.1
 description: GraphQL tools for parsing, transforming and printing GraphQL documents.
 repository: https://github.com/gql-dart/gql
 environment: 
--- a/gql/CHANGELOG.md
+++ b/gql/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.15.0
+## 0.14.1
 
 - Add a const constructor to `Visitor` and the applicable sub-classes.
 


ValentinVignal:
`Dart / Check gql_example_cli_github (pull_request)` fails on the package `gql_example_cli_github` with

```
A list of incorrectly formatted files may follow:

 multipack  gql_example_cli_github 
 multipack  dart format . --set-exit-if-changed 
Formatted lib/__generated__/schema.schema.gql.g.dart
Formatted 31 files (1 changed) in 4.49 seconds.
 multipack  failure 1
```

But if I run

```zsh
dart format .
```

I have no changes on my local. I'm not sure where it comes from. Do you have an idea?

knaeckeKami:
fixed it in https://github.com/gql-dart/gql/pull/396 , I had to pin an older version of dart_style, so that the code generation uses the same version of dart formatting as the bundled `dart format` command does.

could you update your branch so CI passes?

ValentinVignal:
Thank you @knaeckeKami !

gql/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: gql
-version: 0.14.0
+version: 0.15.0
```
knaeckeKami, line 2: Since we are still at version <1.0.0, a bump from 0.14.0 to 0.15.0 would indicate a breaking change like a major version bump according to 
[pub_semver](https://pub.dev/packages/pub_semver).

Since gql is used by [many packages](https://pub.dev/packages?q=dependency%3Agql), major version bumps like this cause a lot of pain for user, so I would like to avoid them when possible.

Since these changes are backwards compatible, I'd prefer to just bump to 0.14.1.  

I think it's time for many packages of the gql ecosystem to be moved to 1.0 version soon, and I hope I'll find time to do that soon, so discussion like this can be avoided in the future :)

PR #403, opened 2023-05-31
Add support for serialize_nulls build config option

twklessor:
Add support for serialize_nulls build config option. This option will enable that null fields are always serialized as null. Currently null fields are serialized at all.


knaeckeKami:
Thanks! I'll take a look at it soon.

knaeckeKami:
So, to get on the same page:

You want a feature that always includes all null values when serialization, right?

Afaik this can be done in a much more simple way by just adding 

`@BuiltValueSerializer(serializeNulls: true)` 

as an annotation to the serializer, no custom serializer is needed for that. The Custom Serializer was needed to have the distinction between absent and null values.

twklessor:
Yes that's right. I almost expected you to have an easier solution. To be honest I used the work you'd already done in the branch with the `Value` wrapper concept.

twklessor:
Sorry to come back months later but what is the status of this?

PR #404, opened 2023-06-03
Remove deprecated libraries from README

ueman:
This removes a couple of deprecated third party libraries from the readme


knaeckeKami:
Thanks!

PR #408, opened 2023-07-07
allow cancellation of call using dio cancellation

rgplvr:
allow cancellation of call using dio cancellation


knaeckeKami:
Hello, thanks for the PR. It seems there is an issue with the CI, looks like a breaking change in one of the dependencies of multipack. I'll try to fix it.

Support for cancellation in Dio is good, but is passing a signal cancel token for all requests of the link the right approach? 

You can only cancel the token once, right?

I think a better solution would be to pass the cancel token via the context. Then the client code could either create a new cancel token for each request or re-use the same token for multiple requests if preferred.


See e.g. https://github.com/gql-dart/gql/blob/master/links/gql_dio_link/lib/src/dio_link.dart#L312


```dart

class DioLinkCancelTokenContextEntry extends ContextEntry {

     final CancelToken token;

     ...
}
```
in DioLink:


```dart

...
final cancelToken = request.context.entry<DioLinkCancelTokenContextEntry>();

...

```

rgplvr:
Makes sense. I have made the changes to push the cancellation token to the context entry.

knaeckeKami:
please update your branch from master to fix the ci issues

rgplvr:
done

links/gql_dio_link/lib/src/dio_link.dart
```diff
@@ -30,6 +31,7 @@ extension _CastDioResponse on dio.Response {
 class DioLink extends Link {
   /// Endpoint of the GraphQL service
   final String endpoint;
+  // final dio.CancelToken? cancelToken;
```
knaeckeKami, line 34: pls delete commented leftover code

links/gql_dio_link/lib/src/dio_link.dart
```diff
@@ -66,8 +68,7 @@ class DioLink extends Link {
 
   @override
   Stream<Response> request(Request request, [forward]) async* {
-    final dio.Response<Map<String, dynamic>> dioResponse =
-        await _executeDioRequest(
+    final dio.Response<Map<String, dynamic>> dioResponse = await _executeDioRequest(
```
knaeckeKami, line 71: pls run `dart format`

knaeckeKami (changes requested):
please delete the commented out code and run dart format

rgplvr:
made the changes the source is properly formatted as per the diff

knaeckeKami (approved):

knaeckeKami:
Thanks, LGTM

PR #412, opened 2023-08-09
[gql_websocket_link] enhancement for websocket connection errors

dehypnosis:
## Background

Hi Team,

This PR is recreated from  https://github.com/gql-dart/gql/pull/388 due to lack of feedbacks and a branch issue from my side. Please check original threads because there were some feedbacks and patch history for that (@agent3bood @knaeckeKami).

Thank you for reviewing it.

### 1. Subscription stream become broken when got WebSocketChannelException.

Currently Subscription request is losing message streams when the client got WebSocketChannelException.
And this problem is cannot be recovered automatically even if autoReconnect is set true, so the websocket connection recovered later. It means when client have a connection problem, all message stream subscriptions for 'Subscription Operation' should be manually recreated.

Fixed this issue by not closing message subscription on WebSocketChannelException.

### 2. No way to get ConnectionError directly from application.

Added StreamController<ConnectionError> _connectionErrorController and public accessor connectionErrorStream.

A practical use case is that, when web socket connection failed due to server side decision or unexpected error, client can be notified about detail of error to take some action.

For example from my case, I need to refresh my client-side id token when it has been expired. I am using Firebase Authentication, and I can make Firebase detect token expiration on App startup. But when app became active from background after more than one hour, I should be detect token is expired or revoked by server.

I made this patch to deal with above case, or any other server-side connection errors.


knaeckeKami:
@agent3bood are you able to triage this? I'm not actively using the websocket so I cannot say too much about this

agent3bood:
@dehypnosis Thanks for the second PR, I ned to take a closer look at it. 
By reading our current code I can see other possible improvements to be made.

Issue #415, opened 2023-08-30
OperationException(linkException: ResponseFormatException(originalException: type 'Null' is not a subtype of type 'String' in type cast 


Routee:
![image](https://github.com/gql-dart/gql/assets/24647390/10bfb1f1-1ab2-4ebc-802a-be9bf56751d8)

if the backend throw a custom exception which message is null. This line: 
```dart 
error['message'] as String
```
 will throw an excption. it's not the expected.

knaeckeKami:
Actually, a message is required by the graphql spec

See https://spec.graphql.org/October2021/#sec-Errors.Error-result-format

PR #416, opened 2023-08-30
fix issue415

Routee:
[!Issue 415](https://github.com/gql-dart/gql/issues/415)


knaeckeKami:
Hi! 

Thanks for you PR. 

The GraphQL spec actually required every error to have a String `message`, see

https://spec.graphql.org/October2021/#sec-Errors.Error-result-format

So I think throwing an Exception is the right behaviour, even though the Exception could be more descriptive I admit.

If you want non-standard behavior like this, you can pass your own ResponseParser to your HttpLink/DioLink.

PR #413, opened 2023-08-09
[gql_code_builder] optimize slow code generation to reduce build time as 10% of previous.

dehypnosis:
## Background

Hi Team and especially @knaeckeKami,

There is an old issue in this library. Which is gql-dart/ferry#143, first reported at Jan, 2021.
I made this PR to address this issue again and share the working draft of a possible solution.

There can be many aspects making build slower, but here I focused on how `gql_code_builder` generates data models.
A code generation for GraphQL type class and built value class gets extremely slower as schema and query gets larger.

The main reasons for that I found:
- With **recursively nesting fragments**, the number of models generated grows exponentially.
- Even for a **single FragmentSpreadNode selection** that is identical to a fragment model already created, a builder will generate the same code just with a different name.

## Example
```graphql
query TestQuery {
    currentUser {
        ...UserFragment
    }

    currentUser2: currentUser {
        ...UserFragment
        location {
            lat
            lng
        }
        moderation {
            ratingPlaceholder
        }
    }
}

fragment UserFragment on User {
   # ...OMITTED
   location {
      ...LocationFragment
   }
   moderation {
      ...ModerationFragment
   }
}

# ...OMITTED
```

Assume we have above query and fragments (though it seems strange).

- Here we can replace `GUserFragmentData_location` to `GLocationFragmentData`. Also for moderation field too.
- And, we know that we can reuse `GUserFragmentData` for `GTestQueryData_currentUser`.
- Furthermore, if `UserFragment` already includes both `location {lat, lng}` and `moderation {ratingPlaceholder}`, here `GUserFragmentData` can be used for  `GTestQueryData_currentUser2` too.

Because of this graph-way nature of GraphQL, we can reuse a lot of our model code. However, the current code generator does not reflect this reuse logic, and so recursively nested Fragments are generating a huge amount of similar code.

## Changes
I created this draft solution that performs the optimizations for **reusable fragment models** mentioned in the example above, including **InlineFragmentNode and WhenExtension support** too. So it covers all models generated from current builder.

I said it is a draft because if a builder with this patch generates code against an existing schema, it will not generate many of the models that were previously generated, and it will also change the type signatures of the reusable getters in many of the former models. This could be a breaking change for users, so for now, I look forward to discussing this approach further.

And there has been made **no regression tests** for this patch yet.
But I am using this patched builder with my own product and found no issues for now.

<details>
<summary>See, there were breaking changes as mentioned above, so I had to fix few type signatures and model names.</summary>

[...]

PR #417, opened 2023-09-01
[gql_code_builder ] bugfix for reuse-fragments feature

dehypnosis:
This PR is a succession of https://github.com/gql-dart/gql/pull/413
After the above PR, our team have developed our ongoing project with dev release version which includes above PR.
And we found a improvement point and a bug to make the reuse-fragment feature stable.
So here made a PR to finalize this feature.

## Changes
- Fix inline fragments data class serialization runtime error with reuse-fragments option enabled.
- Fix WhenExtension to not to create redundant mappers for interfaces which doesn't make sense because interfaces don't have __typename. (But for the backward compatibility, this change is only applied when reuseFragment option is enabled.)

### DIFF
--- a/codegen/gql_code_builder/lib/data.dart
+++ b/codegen/gql_code_builder/lib/data.dart
@@ -27,7 +27,7 @@ Library buildDataLibrary(
   ),
 ]) {
   final fragmentMap = _fragmentMap(docSource);
-  final possibleTypesMap = _possibleTypesMap(schemaSource);
+  final possibleTypesMap = dataClassConfig.reuseFragments ? _possibleTypesMap(schemaSource) : <String, Set<String>>{};
   final dataClassAliasMap = dataClassConfig.reuseFragments
       ? _dataClassAliasMap(docSource, fragmentMap, possibleTypesMap)
       : <String, Reference>{};
@@ -75,13 +75,13 @@ Library buildDataLibrary(
 }
 
 Map<String, SourceSelections> _fragmentMap(SourceNode source) => {
-      for (var def
+      for (final def
           in source.document.definitions.whereType<FragmentDefinitionNode>())
         def.name.value: SourceSelections(
           url: source.url,
           selections: def.selectionSet.selections,
         ),
-      for (var import in source.imports) ..._fragmentMap(import)
+      for (final import in source.imports) ..._fragmentMap(import)
     };
 
 Map<String, Set<String>> _possibleTypesMap(SourceNode source,
--- a/codegen/gql_code_builder/lib/data.dart
+++ b/codegen/gql_code_builder/lib/data.dart
@@ -27,7 +27,9 @@ Library buildDataLibrary(
   ),
 ]) {
   final fragmentMap = _fragmentMap(docSource);
-  final possibleTypesMap = dataClassConfig.reuseFragments ? _possibleTypesMap(schemaSource) : <String, Set<String>>{};
+  final possibleTypesMap = dataClassConfig.reuseFragments
+      ? _possibleTypesMap(schemaSource)
+      : <String, Set<String>>{};
   final dataClassAliasMap = dataClassConfig.reuseFragments
       ? _dataClassAliasMap(docSource, fragmentMap, possibleTypesMap)
       : <String, Reference>{};
--- a/codegen/gql_code_builder/lib/data.dart
+++ b/codegen/gql_code_builder/lib/data.dart
@@ -4,6 +4,7 @@ import "package:gql/ast.dart";
 import "package:gql_code_builder/src/common.dart";
 import "package:gql_code_builder/src/config/data_class_config.dart";
 import "package:gql_code_builder/src/config/when_extension_config.dart";
+import "package:gql_code_builder/src/utils/possible_types.dart";
 
 import "./source.dart";
 import "./src/operation/data.dart";
@@ -26,8 +27,9 @@ Library buildDataLibrary(
   ),
 ]) {
   final fragmentMap = _fragmentMap(docSource);
+  final possibleTypesMap = _possibleTypesMap(schemaSource);
   final dataClassAliasMap = dataClassConfig.reuseFragments
-      ? _dataClassAliasMap(docSource, fragmentMap)
+      ? _dataClassAliasMap(docSource, fragmentMap, possibleTypesMap)
       : <String, Reference>{};
 
   final operationDataClasses = docSource.document.definitions
@@ -40,6 +42,7 @@ Library buildDataLibrary(
           typeOverrides,
           whenExtensionConfig,
           fragmentMap,
+          possibleTypesMap,
           dataClassAliasMap,
         ),
       )
@@ -55,6 +58,7 @@ Library buildDataLibrary(
           typeOverrides,
           whenExtensionConfig,
           fragmentMap,
+          possibleTypesMap,
           dataClassAliasMap,
         ),
       )
@@ -80,16 +84,40 @@ Map<String, SourceSelections> _fragmentMap(SourceNode source) => {
       for (var import in source.imports) ..._fragmentMap(import)
     };
 
+Map<String, Set<String>> _possibleTypesMap(SourceNode source,
+    [Set<String>? visitedSource, Map<String, Set<String>>? possibleTypesMap]) {
+  visitedSource ??= {};
+  possibleTypesMap ??= {};
+
+  source.imports.forEach((s) {
+    if (!visitedSource!.contains(source.url)) {
+      visitedSource.add(source.url);
+      _possibleTypesMap(s, visitedSource, possibleTypesMap);
+    }
+  });
+
+  source.document.possibleTypesMap().forEach((key, value) {
+    possibleTypesMap![key] ??= {};
+    possibleTypesMap[key]!.addAll(value);
+  });
+
+  return possibleTypesMap;
+}
+
 Map<String, Reference> _dataClassAliasMap(
-    SourceNode source, Map<String, SourceSelections> fragmentMap,
-    [Map<String, Reference>? aliasMap, Set<String>? visitedSource]) {
+    SourceNode source,
+    Map<String, SourceSelections> fragmentMap,
+    Map<String, Set<String>> possibleTypesMap,
+    [Map<String, Reference>? aliasMap,
+    Set<String>? visitedSource]) {
   aliasMap ??= {};
   visitedSource ??= {};
 
   source.imports.forEach((s) {
     if (!visitedSource!.contains(source.url)) {
       visitedSource.add(source.url);
-      _dataClassAliasMap(s, fragmentMap, aliasMap);
+      _dataClassAliasMap(
+          s, fragmentMap, possibleTypesMap, aliasMap, visitedSource);
     }
   });
 
@@ -101,6 +129,7 @@ Map<String, Reference> _dataClassAliasMap(
       selections: def.selectionSet.selections,
       fragmentMap: fragmentMap,
       aliasMap: aliasMap,
+      possibleTypesMap: possibleTypesMap,
     );
   }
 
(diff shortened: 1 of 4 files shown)


knaeckeKami:
Thanks, I'll have a proper look later.

One thing that i noticed on the first glance, you seem to have implemented another way of calculating the possibleTypes map.

Would it be feasible to leverage https://github.com/gql-dart/gql/blob/master/codegen/gql_code_builder/lib/src/utils/possible_types.dart instead ?

codegen/gql_code_builder/lib/data.dart
```diff
@@ -71,25 +77,49 @@ Library buildDataLibrary(
...
+    if (!visitedSource!.contains(source.url)) {
+      visitedSource.add(source.url);
+      _possibleTypesMap(s, visitedSource, possibleTypesMap);
+    }
+  });
+
+  source.document.possibleTypesMap().forEach((key, value) {
+    possibleTypesMap![key] ??= {};
+    possibleTypesMap[key]!.addAll(value);
+  });
```
dehypnosis, line 104: I think this is already utilizing what you mentioned. Isn't it?
But additionally I made a recursion to make a global possibleTypes map thorough all imported graphql files.

knaeckeKami: yes, sorry missed that.

knaeckeKami:
sorry for the late response! 
please pull the latest changes from master to ci succeeds

dehypnosis:
@knaeckeKami  how are you doing well? just merged upstream/master branch.

knaeckeKami:
Thanks! It seems there are now difference in the generated code in end_to_end_test, despite the reuse_fragments flag being disabled. 
I'll look into why this is.

dehypnosis:
@knaeckeKami oh, sorry for unexpected trouble. Is that related to WhenExtension ?

codegen/gql_code_builder/lib/src/inline_fragment_classes.dart
```diff
@@ -94,28 +97,34 @@ List<Spec> buildInlineFragmentClasses({
...
+        dataClassAliasMap: dataClassAliasMap,
+        schemaSource: schemaSource,
+        type: inlineFragment.typeCondition!.on.name.value,
+        typeOverrides: typeOverrides,
+        superclassSelections: {
+          name: SourceSelections(url: null, selections: selections),
+          ...Map.fromEntries(superclassSelections.entries.map((e) => MapEntry(
+              "${e.key}__as${inlineFragment.typeCondition!.on.name.value}",
+              e.value))),
+        },
```
dehypnosis, line 124: @knaeckeKami  I think the cause is that this fix about `buildInlineFragmentClasses` is not checking feature flag.
This is made to cover more interfaces like adding `GheroFieldsFragment__asHuman` interface to `GheroFieldsFragmentData__asHuman`. And I think changes made by this are all additions for missing interfaces but no regressions of existing interfaces. So there would be no regression issue even without the feature flag.

codegen/gql_code_builder/lib/src/when_extension.dart
```diff
@@ -40,9 +40,15 @@ Extension? inlineFragmentWhenExtension(
...
+    required Map<String, Set<String>> possibleTypesMap,
     required Map<String, Reference> dataClassAliasMap}) {
   final inlineFragmentsWithTypConditions = inlineFragments
-      .where((inlineFragment) => inlineFragment.typeCondition != null)
+      .where((inlineFragment) =>
+          inlineFragment.typeCondition != null
+          // except interfaces on when extension
+          &&
+          !possibleTypesMap
+              .containsKey(inlineFragment.typeCondition!.on.name.value))
```
dehypnosis, line 51: @knaeckeKami And I just checked WhenExtension codes too, And it seems okay. There would be no changes without feature flag, because added condition [line 49-51] will be always true when the flag is off.

codegen/gql_code_builder/lib/src/inline_fragment_classes.dart
knaeckeKami: I just run the code generation without the flag and unfortunately, it produces code that does not compile.

the following graphql document caues problems:

```graphql
query HeroWithInterfaceSubTypedFragments($episode: Episode!) {
  hero(episode: $episode) {
    ...heroFieldsFragment
  }
}

fragment heroFieldsFragment on Character {
  id
  name

  ...on Human {
    ...humanFieldsFragment
  }

  ...on Droid {
    ...droidFieldsFragment
  }
}

fragment humanFieldsFragment on Human {
  homePlanet
  friends {
    ...on Droid {
      id
      name
      ...droidFieldsFragment
    }
    ...on Human {
      id
      name
      homePlanet
    }
  }
}

fragment droidFieldsFragment on Droid {
  primaryFunction
}

```

this adds another superclass to GheroFieldsFragmentData__asHuman which conflicts with the implementation:


```
'GheroFieldsFragmentData__asHuman.friends' ('BuiltList<GheroFieldsFragmentData__asHuman_friends?>? Function()') isn't a valid override of 'GheroFieldsFragment__asHuman.friends' ('BuiltList<GheroFieldsFragment__asHuman_friends?>? Function()'). (Documentation) 
The member being overridden (hero_with_interface_subtyped_fragments.data.gql.dart:151).
```

dehypnosis: @knaeckeKami sorry to bother you, I will turn off new logics when there is no feature flag. Will share again soon.

knaeckeKami: no worries! i really appreciate you working on this! :)

dehypnosis: @knaeckeKami
How are you doing well? Finally came back here.
Sorry for late response.  Just made the update. 😀

knaeckeKami:
Thanks! 

A user reported some issues with unions with a lot of possible values, see here:

https://github.com/LiLatee/flutter_ferry_unions_issue

I noticed that the DFS searches in all permutations of the given union types, which leads to exponential code generation time . 

This is a possible fix:

https://github.com/gql-dart/gql/tree/possible_reuse_fragment_fix


Please add a test that checks if this bug is fixed by this PR

LiLatee:
Hi @knaeckeKami @dehypnosis! 😄 

I've prepared a [repo](https://github.com/LiLatee/flutter_ferry_unions_issue/tree/main/flutter_ferry_unions_issue2) with an example that generates a code with errors. In short, overridden fields have a wrong return type. I think that the problem exists because I have two different fragments defined on a single type defined in schema. In my case I have two fragments:: 
`ActivityB` and `ActivityBWithoutSharedItem` defined on type `ActivityB`.

Is there a quick solution for that? 🤔

LiLatee:
I've spent a few more hours and created an even much simpler [example](https://github.com/LiLatee/flutter_ferry_unions_issue/tree/main/flutter_ferry_unions_issue2) to reproduce the issue and I really cannot understand what's the problem here. 
image

I've found three solutions to fix the issue by modifying fragments:
1. Remove import. Just import (even without using it) causes the error.
image

2. Remove one fragment and use its content directly. Removing `ItemEdge` here and pasting that fragment's content fixes the issue
image
image

3. Or removing one union field `activityData`
image

Each of these modifications fixes the issue and I need to use just one of them.

Is there a chance that you could take a look at that in the near future? 🙏

knaeckeKami:
Hi!

I also looked a bit into it.
I think the issue even occurs in the end_to_end_test in this query:

`hero_with_interface_subtyped_fragments.graphql`.

```graphql
query HeroWithInterfaceSubTypedFragments($episode: Episode!) {
  hero(episode: $episode) {
    ...heroFieldsFragment
  }
}

fragment heroFieldsFragment on Character {
  id
  name

  ...on Human {
    ...humanFieldsFragment
  }

  ...on Droid {
    ...droidFieldsFragment
  }
}

fragment humanFieldsFragment on Human {
  homePlanet
  friends {
    ...on Droid {
      id
      name
      ...droidFieldsFragment
    }
    ...on Human {
      id
      name
      homePlanet
    }
  }
}

fragment droidFieldsFragment on Droid {
  primaryFunction
}

```

And it seems to be triggered by Inline fragment spreads with typeconditon where the selection is just one fragment spread. 

But I don't know much more yet.

knaeckeKami:
I now understand what the issue is. 

I found a way to fix it in gql_code_builder, but not sure if this breaks smth else. 

@LiLatee can you try

```dependency_overrides:
gql_code_builder:
   git:
       url: https://github.com/gql-dart/gql/tree/repro_reuse_fragments
       ref: repro_reuse_fragments
       path: codegen/gql_code_builder
```

and give me feedback?

LiLatee:
Unfortunately still the same issue on the example [repository](https://github.com/LiLatee/flutter_ferry_unions_issue/tree/main/flutter_ferry_unions_issue2) 😞 
image

knaeckeKami:
yes, thanks for fixing the broken link.

Sorry. I did not push all the code. please try again.

LiLatee:
I am getting other errors right now
image

I can add `Data` to the name, but not sure if it's fine. 
image

I can check it tomorrow

knaeckeKami:
Ok, I had a suspicion that it would be not that easy ;)

Would appreciate if you could create failing reproduction to the end_to_end_test repo.

But I know now what the issue is, it's only and I think I can find a fix in the near future.

LiLatee:
> Would appreciate if you could create failing reproduction to the end_to_end_test repo.

Sure, I can try to take care of it tomorrow 👌

> But I know now what the issue is, it's only and I think I can find a fix in the near future.

Would be awesome 🤞

codegen/gql_code_builder/lib/src/inline_fragment_classes.dart
```diff
@@ -94,28 +97,35 @@ List<Spec> buildInlineFragmentClasses({
...
+        ),
+        fragmentMap: fragmentMap,
+        possibleTypesMap: possibleTypesMap,
+        dataClassAliasMap: dataClassAliasMap,
+        schemaSource: schemaSource,
+        type: inlineFragment.typeCondition!.on.name.value,
+        typeOverrides: typeOverrides,
+        superclassSelections: {
+          name: SourceSelections(url: null, selections: selections),
+          if (dataClassAliasMap.isNotEmpty)
```
knaeckeKami, line 121: I think this condition is not correct.
Inline Fragments can add new fields in selections which can cause conflicts in extending the fragments. 

It also causes the build error in the end to end test package.

When Removing this, it works again

knaeckeKami (changes requested):
U think that's the issue

LiLatee:
> Would appreciate if you could create failing reproduction to the end_to_end_test repo.

@knaeckeKami should I create a failing case for the code before you commented that condition above?
Because your comment removes some part of fragment reusing. E.g. before that change my `user.data.gql.dart` file had 212 lines and now it has 712, because new fragments (that already exist) are created again.

LiLatee:
> Would appreciate if you could create failing reproduction to the end_to_end_test repo.

@knaeckeKami should I create a failing case for the code before you commented that condition above?
Because your comment removes some part of fragment reusing. E.g. before that change my `user.data.gql.dart` file had 212 lines and now it has 712, because new fragments (that already exist) are created again.

knaeckeKami:
Failing in the sense that it's more code or that the code doesn't compile?

Anyway, yes please share.

I know that this might increase the code size but this optimization is not correct in the way it is now. But maybe we can find I way to make it work again.

dehypnosis:
Sorry guys, I'm busy doing my day works nowadays. I will dive into to take a look in this weekend if the issue is not resolved until then.

LiLatee:
@knaeckeKami oh I've finally found one issue. It looks like the order of the imports in .graphql files is important.
e.g.
Case 1:
image
image

Case 2:
image
image

knaeckeKami:
thanks! 

@LiLatee implemented another approach to reuse fragment classes in #437 . this is just an experiment at this point, but feel free to try it out.

dav3rin:
Hi 

Does anyone have a working fix for reuse-fragments feature?

When I set `reuse_fragments` to true, build-runner freezes. This happens on both this commit and the latest commit.

Issue #411, opened 2023-08-05
[graphql_transport_ws] The new subprotocol doesn't auto-reconnect


agufagit:
When debugging, when app is hot-reloaded, it throws an error

```
I/flutter ( 6092): ResponseFormatException(originalException: type '(dynamic) => Map<String, dynamic>?' is not a subtype of type '(dynamic) => FutureOr<Map<String, dynamic>>?' of 'function result', originalStackTrace: #0      TransportWsClientOptions.graphQLSocketMessageDecoder (package:gql_websocket_link/src/graphql_transport_ws.dart)
graphql_transport_ws.dart:1
I/flutter ( 6092): #1      _ConnectionState._startConnecting.<anonymous closure>.<anonymous closure>
graphql_transport_ws.dart:778
I/flutter ( 6092): #2      _RootZone.runUnaryGuarded (dart:async/zone.dart:1594:10)

```

I expect it to auto-reconnect.  The old protocol (gql_websocket_link.dart) works fine, but the new subprotocol throws this error.

file: graphql_transport_ws.dart

line: 778

juancastillo0:
Are you providing a custom `graphQLSocketMessageDecoder`?

It seems the Function is not properly casted, maybe a solution could be to put the `?` inside the `FutureOr` in `GraphQLSocketMessageDecoder`

Current:
https://github.com/gql-dart/gql/blob/e021a3774f77701a534c427d702df4b7781c4a66/links/gql_websocket_link/lib/src/link.dart#L16-L17

Proposed Solution:
```dart
typedef GraphQLSocketMessageDecoder = FutureOr<Map<String, dynamic>?> Function( 
     dynamic message); 
```

Could you try providing an anonymous callback to the options configuration? 

```dart
TransportWsClientOptions(
  graphQLSocketMessageDecoder: (dynamic message) =>
      jsonDecode(message as String) as FutureOr<Map<String, dynamic>>?
)
```

Thanks!

agufagit:
I'm not providing a custom graphQLSocketMessageDecoder.

The anonymous callback seems to work initially, it's not throwing error anymore.  Not sure how it works with data though.

I'll close this issue after some integration tests in a few days.

Thank you

PR #418, opened 2023-09-14
close reason should be optional and options.shouldRetry -> bool as a replacement for isFatalConnectionProblem

catapop84:
This PR fixes 2 problems

1. Server can close connection without providing any reason. With current implementation this will throw an error because ` reason: socket.closeReason!` is expected to exists. 
2. Since the javascript library implementation is the main reference, they introduce shouldRetry: boolean option and deprecated `isFatalConnectionProblem`

Based on their documentation: https://github.com/enisdenjo/graphql-ws/blob/master/src/client.ts , shouldRetry can override fatal close code which is the default implementation.

```
   * The library classifies the following close events as fatal:
   * - _All internal WebSocket fatal close codes (check `isFatalInternalCloseCode` in `src/client.ts` for exact list)_
   * - `4500: Internal server error`
   * - `4005: Internal client error`
   * - `4400: Bad request`
   * - `4004: Bad response`
   * - `4401: Unauthorized` _tried subscribing before connect ack_
   * - `4406: Subprotocol not acceptable`
   * - `4409: Subscriber for <id> already exists` _distinction is very important_
   * - `4429: Too many initialisation requests`
   *
   * In addition to the aforementioned close events, any _non-CloseEvent_ connection problem
   * is considered fatal by default. However, this specific behaviour can be altered by using
   * the `shouldRetry` option.
```

### DIFF
--- a/codegen/end_to_end_test/lib/aliases/__generated__/alias_var_fragment.data.gql.dart
+++ b/codegen/end_to_end_test/lib/aliases/__generated__/alias_var_fragment.data.gql.dart
@@ -17,14 +17,17 @@ abstract class GPostsData implements Built<GPostsData, GPostsDataBuilder> {
 
   static void _initializeBuilder(GPostsDataBuilder b) =>
       b..G__typename = 'Query';
+
   @BuiltValueField(wireName: '__typename')
   String get G__typename;
   BuiltList<GPostsData_posts?>? get posts;
   static Serializer<GPostsData> get serializer => _$gPostsDataSerializer;
+
   Map<String, dynamic> toJson() => (_i1.serializers.serializeWith(
         GPostsData.serializer,
         this,
       ) as Map<String, dynamic>);
+
   static GPostsData? fromJson(Map<String, dynamic> json) =>
       _i1.serializers.deserializeWith(
         GPostsData.serializer,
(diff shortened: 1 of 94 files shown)


catapop84:
Sometimes we make updates to our server and we need to reset the program. On closing server sends close code 1002 and no reason. 
This way we can override if we choose to to reconnect once the server is rebooted. This will also make the library in line with the javascript equivalent .

PS. I know this might be confusing, but this PR is for the new websocket protocol: `graphql-transport-ws` from the library `graphql-ws `

knaeckeKami:
@juancastillo0 what do you think?

juancastillo0:
Hi! Looks great! The implementation is a little different form the reference, but the differences make sense. There are no additional tests, however the changes are kind of incremental.

Not sure if we should treat it as a breaking change, since the close reason was non-nullable. We could also make it non-null and pass an empty String (or the close code as a String) when the server does not return a reason, but I personally don't like it. And I suppose it would still be a breaking change since the `TransportWsClientOptions` has a new field and it could be implemented, but I believe it would be less common.

If we make a breaking change maybe we could accommodate other changes like this one https://github.com/gql-dart/gql/issues/411#issuecomment-1666551422 and I am not sure what is the status of this PR https://github.com/gql-dart/gql/pull/412.

In general I think it could be merge, not sure what you think about the test and the breaking change.

knaeckeKami:
@catapop84 please pull the latest changes from master to CI succeeds

catapop84:
@knaeckeKami done

catapop84:
@knaeckeKami can you try one more time. Should be valid this time.

PR #424, opened 2023-10-27
[gql_example_http_auth_link] add request queuing while refreshing token

2shrestha22:
This PR will queue incoming requests during the token refresh process. The token refresh queue ensures that only one token refresh occurs at a time, preventing multiple concurrent refreshes.


knaeckeKami:
thanks!

PR #429, opened 2023-11-17
feat(dedupe_link): Add shouldDedupe option

nhannah:
Dedupe link is an all or nothing solution at the moment, but in use it is likely you would want to dedupe queries, but not mutations, or possibly filter deduplication in some other granular way. This modification allows for devs to consider the incoming request and prevent deduplication if desired.

### DIFF
--- a/links/gql_dedupe_link/lib/gql_dedupe_link.dart
+++ b/links/gql_dedupe_link/lib/gql_dedupe_link.dart
@@ -2,28 +2,37 @@
 library gql_dedupe_link;
 
 import "package:async/async.dart";
+import "package:gql/ast.dart";
 import "package:gql_exec/gql_exec.dart";
 import "package:gql_link/gql_link.dart";
 
 /// A [Link] to deduplicate [Request]s
 class DedupeLink extends Link {
+  final List<OperationType> _nonDedupeTypes;
   final Map<Request, StreamSplitter<Response>> _inFlight = {};
+
+  DedupeLink({List<OperationType> nonDedupeTypes = const []})
+      : _nonDedupeTypes = nonDedupeTypes;
 
   @override
   Stream<Response> request(
     Request request, [
     NextLink? forward,
   ]) {
-    if (_inFlight.containsKey(request)) {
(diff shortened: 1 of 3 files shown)


knaeckeKami:
Thanks!

Do you think it would make sense to pass in a `bool Function shouldConsiderForDeduplication(Request request)` instead of a list of operation types?

the user could then pass in  `(req) => req.operation.getOperationType() == OperationType.query`

nhannah:
> Thanks!
> 
> Do you think it would make sense to pass in a `bool Function shouldConsiderForDeduplication(Request request)` instead of a list of operation types?
> 
> the user could then pass in `(req) => req.operation.getOperationType() == OperationType.query`

Always a fan of allowing more freedom, discussed with a colleague and went with `shouldPreventDedupe` so logic written by a developer can be inclusive of the group they want to prevent deduplication on, i.e.  `(req) => req.operation.getOperationType() == OperationType.mutation` vs `(req) => req.operation.getOperationType() != OperationType.mutation`

knaeckeKami:
Thanks! one last small nit:

From Effective Dart:

> PREFER the “positive” name for a boolean property or variable
Most boolean names have conceptually “positive” and “negative” forms where the former feels like the fundamental concept and the latter is its negation—”open” and “closed”, “enabled” and “disabled”, etc. Often the latter name literally has a prefix that negates the former: “visible” and “in-visible”, “connected” and “dis-connected”, “zero” and “non-zero”.
When choosing which of the two cases that true represents—and thus which case the property is named for—prefer the positive or more fundamental one. Boolean members are often nested inside logical expressions, including negation operators. If your property itself reads like a negation, it’s harder for the reader to mentally perform the double negation and understand what the code means.

https://dart.dev/effective-dart/design#prefer-the-positive-name-for-a-boolean-property-or-variable)

`shouldPreventDedupe` is IMO conceptually slightly more complicated than `shouldDedupe`.

I would prefer a positively worded field (which then would default to true to dedupe every request, which is the current behaviour)

nhannah:
Works for me I'll swap it around and update.

PR #437, opened 2024-01-07
reuse classed by selection set

knaeckeKami:
- Revert "[gql_code_builder] add dataClassConfig option to reuse class definitions for field selections that include only a single inline fragment spread"
- simpler version of reuse fragments with less scope
- simpler version of reuse fragments with less scope
- simpler version of reuse fragments with less scope


dav3rin:
Hi @LiLatee and @knaeckeKami 

I am currently have fatal issue in my codebase:

```
../../../flutter/third_party/dart/runtime/vm/zone.cc: 96: error: Out of memory.
```

My schema is notably quite large. And I believe I have now added the last straw.

Line counts that work:
```
schema.ast.gql.dart -21296
schema.gql.g.dart - 23224
serializers.gql.dart - 18643
serializers.gql.g.dart - 25492
```

Line counts that result in failure:
```
schema.ast.gql.dart - 21296
schema.gql.g.dart - 23224
serializers.gql.dart - 21206
serializers.gql.g.dart - 29498
```

I'm struggling right now to find a good solution. Do you have any ideas on what I can do in the scenario? Even a temporary fix at this point would be helpful.

Thanks

dav3rin:
Hey @LiLatee 

Thanks for the quick reply.

I did a build and it seems to work! File sizes down and build time went from 12 mins to 2 mins. 

As you predicted, I am missing a few classes. 

Could you drop a link to the python script?

PR #438, opened 2024-01-08
(fix/reuse-fragments) New test case for issue with implementing not existing classes

LiLatee:
@knaeckeKami as you [wanted](https://github.com/gql-dart/gql/pull/417#issuecomment-1879767253) I added a new test case that doesn't work in my case. 

I am not sure if it's done as you prefer and I was not able to test it using that repository, because as I saw it doesn't use local versions of packages and instead it uses versions from pub.dev and unfortunately I haven't found a quick way to overcome it 😅


knaeckeKami:
you can use 

`dart pub global activate multipack`

`multipack pubspec override`  to use the local packages,

and then when you are done 

`multipack pubspec clean` before committing

LiLatee:
Oh, that's something new to me and It works. Thanks! ❤️  But I see that my case doesn't produce an error right now. So I have to dive into it again. I will let you know when it's ready.

knaeckeKami:
probably because end_to_end_test is not using reuse_fragments right now.

LiLatee:
>  But I see that my case doesn't produce an error right now

My mistake... I've created a branch using  https://github.com/gql-dart/gql/pull/417 instead of your https://github.com/gql-dart/gql/tree/repro_reuse_fragments

I ca create a proper one tomorrow 😞 

but it was also true :D 
> probably because end_to_end_test is not using reuse_fragments right now.

LiLatee:
@knaeckeKami now it's fine. I mean now after enabling reuse_fragments `connection.data.gql.dart` will use not existing class.

knaeckeKami:
Thanks! it will probably show that error on #417 though, which master does not have yet. but that will it make easier figure it out

knaeckeKami:
For now, it's sufficient to run it with reuse_fragments disabled.

When we officially ship this, we can think about proper test coverage of code gen with and without reuse_fragment

LiLatee:
@knaeckeKami done 👌

PR #440, opened 2024-01-11
Update uuid package

### LINKED ISSUE #439: [gql_websocket_link] uuid package version conflict
Hello, is it possible to update `uuid` package to ^4.0.0 that the latest [graphql](https://github.com/zino-hofmann/graphql-flutter/tree/main/packages/graphql) package uses & release a new version? Currently, we cannot upgrade `graphql` version because of this.

I know that I can use dependency overriding but I have a modular project, so I have to use overriding in too many places :/

memishood:
Update `uuid` package closes #439


knaeckeKami:
Since the websocket link can support both 3.0.0 and 4.0.0 versions, please change the range to ">=3.0.0 <5.0.0" in order to cause minimal disruption for our users

PR #444, opened 2024-01-26
chore(vars-builder-bugfix): Fixing nullables and type overrides for vars create factories

caffeineflo:
Hi @knaeckeKami!

Unfortunately I found a bug in my previous commit for the vars create factories 🙈 

Nullables were still required and types that were in the typeOverrides, were called with the `toBuilder()` function causing the compiler to fail for these files.

🙈 🙈 🙈 🤦 🤦 🤦 🙈🙈🙈

I fixed it on this branch and hoping we can merge this as a new bugfix version. Sorry for the initial oversight.

As for Ferry: Happy to open a PR making this new bugfix version a dependency if that helps you at all!

Thanks,
Florian!


knaeckeKami:
LGTM! thanks

PR #447, opened 2024-02-16
[gql_code_builder] fix: vars_create_factories handle BuiltList type

LiLatee:
Hi! 👋 
I've tried to use that feature [vars_create_factories](https://github.com/gql-dart/gql/pull/434) and I've encounter an issue with arguments being lists. The parameter of a factory requires a `BuiltList` that is assigned to `ListBuilder` and that is not allowed.
image

@caffeineflo I've found some solution. Not sure whether it's the best way, but it works for me. So if you could take a look 🙏

### DIFF
--- a/codegen/gql_code_builder/lib/src/required_vars_constructor.dart
+++ b/codegen/gql_code_builder/lib/src/required_vars_constructor.dart
@@ -38,12 +38,17 @@ Constructor builtCreateConstructor({
     final typeDefinitionNode = getTypeDefinitionNode(
         schemaSource.document, g.returns!.symbol!.replaceFirst("G", ""));
 
+    final bool isFromBuiltCollectionPackage =
+        g.returns?.url?.contains("built_collection") ?? false;
+
     /// "$BuiltList" is a String "BuiltList<dynamic>" and
     /// [g.returns!.symbol!] returns "BuiltList"
     ///  so we cannot use equality operator here.
-    final isBuiltList = g.returns?.symbol != null
-        ? "$BuiltList".contains(g.returns!.symbol!)
-        : false;
+    final isBuiltList = (g.returns?.symbol != null
+            ? "$BuiltList".contains(g.returns!.symbol!)
+            : false) &&
+        isFromBuiltCollectionPackage;
+
     final isBuiltType = typeDefinitionNode is InputObjectTypeDefinitionNode ||
         typeDefinitionNode is ScalarTypeDefinitionNode ||
         isBuiltList;
--- a/codegen/gql_code_builder/lib/src/required_vars_constructor.dart
+++ b/codegen/gql_code_builder/lib/src/required_vars_constructor.dart
@@ -37,9 +37,16 @@ Constructor builtCreateConstructor({
   final assignments = filteredGetters.map((g) {
     final typeDefinitionNode = getTypeDefinitionNode(
         schemaSource.document, g.returns!.symbol!.replaceFirst("G", ""));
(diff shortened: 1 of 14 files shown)


codegen/gql_code_builder/lib/src/required_vars_constructor.dart
```diff
@@ -37,9 +37,16 @@ Constructor builtCreateConstructor({
   final assignments = filteredGetters.map((g) {
     final typeDefinitionNode = getTypeDefinitionNode(
         schemaSource.document, g.returns!.symbol!.replaceFirst("G", ""));
-    final isBuiltType = typeDefinitionNode is InputObjectTypeDefinitionNode ||
-        typeDefinitionNode is ScalarTypeDefinitionNode;
 
+    /// "$BuiltList" is a String "BuiltList<dynamic>" and
+    /// [g.returns!.symbol!] returns "BuiltList"
+    ///  so we cannot use equality operator here.
+    final isBuiltList = g.returns?.symbol != null
```
knaeckeKami, line 44: can you also check the URL of the reference is from built_value, so we cannot have false positives here?

knaeckeKami (changes requested):
please also add a test with a list input in the end to end test repo so we don't break that again in a refactor

links/gql_websocket_link/lib/src/messages.dart
```diff
@@ -165,7 +165,7 @@ class SubscriptionData extends GraphQLSocketMessage {
   int get hashCode => toJson().hashCode;
 
   @override
-  bool operator ==(dynamic other) =>
+  bool operator ==(Object other) =>
```
caffeineflo, line 168: Is this an unrelated change?

codegen/gql_code_builder/lib/src/required_vars_constructor.dart
caffeineflo: Good call, I thought of doing that too to find `Built` Types at one point (when I wrote some of that code), but I think I didn't end up doing that as it was extremely hard to pin down the correct import. 
For `BuiltList` s though, this would be checking for the built_lists import though, which doesn't change as often, right?


An alternative that might be worth exploring though, would be to check if all `BuiltList` instances are of type `ListType` and `ListType` is exclusively used for `BuiltLists`. Then you could "simply" add another check to L47

knaeckeKami: I think the `url` of the `Reference` should be '"package:built_collection/built_collection.dart", I think this could just be added additionally to this check to avoid clashes with types from the schema, which theoretically also could contain symbols that contain the name `BuiltList`

LiLatee: You meant `built_collection` right? Done ✅

links/gql_websocket_link/lib/src/messages.dart
LiLatee: Yes. To be honest I cannot figure out why but the analyzer was failing 
image

Your commit was the last one and then it was succeeding. Now it doesn't 🤷‍♂️

LiLatee: Maybe because workflow uses `dart:latest`, so something in the Dart could change.
image

knaeckeKami: yes, it's likely a new lint rule in dart

LiLatee: I couldn't find that 😅

PR #449, opened 2024-02-20
#feat: Add value or absent option

LiLatee:
Hey! 👋 
I've been lacking the option to send a value if it exists and not send any value (neither null) if not present.

~~EDIT: Converted to DRAFT, because it's not working yet.~~

### DIFF
--- a/codegen/gql_tristate_value/lib/src/value.dart
+++ b/codegen/gql_tristate_value/lib/src/value.dart
@@ -9,21 +9,24 @@ sealed class Value<T extends Object> {
   /// The value is absent. It will not be serialized.
   const factory Value.absent() = AbsentValue;
 
-  /// If the value is non-null, it will be serialized as the value.
-  /// The value is absent. It will not be serialized.
-  const factory Value.valueOrAbsent(T? value) = ValueOrAbsent;
-
   /// The value is present. It may still be be null.
   /// If the value is null, it will be serialized as null.
   /// If the value is non-null, it will be serialized as the value.
   const factory Value.present(T? value) = PresentValue<T>;
 
+  /// If the value is non-null, it will be serialized as the value.
+  /// If the value is null. It will not be serialized.
+  factory Value.ofNullable(T? value) {
+    if (value == null) {
+      return Value.absent();
+    }
+    return Value.present(value);
+  }
+
   /// Returns the value if present (no matter if null or non-null), otherwise throws a [StateError].
   T? get requireValue => switch (this) {
         PresentValue(:final value) => value,
         AbsentValue() => throw StateError("Value is absent"),
-        ValueOrAbsent(:final value) =>
-          value ?? (throw StateError("Value is absent")),
       };
 
   /// return the value if present and non-null, otherwise null.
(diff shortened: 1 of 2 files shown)


knaeckeKami:
Why do you need a separate class for that?
Adding a new subtype to a sealed class is a breaking change.

And there are no new states - a value is either present or absent. we don't need a separate type for this, this could be a static method or factory, could'nt it?

codegen/gql_tristate_value/test/value_test.dart
```diff
@@ -2,33 +2,107 @@ import 'package:gql_tristate_value/gql_tristate_value.dart';
 import 'package:test/test.dart';
 
 void main() {
-  test("requireValue on absent throws", () {
-    final absent = const AbsentValue();
+  group(
+    'requireValue',
+    () {
+      test('on $AbsentValue throws', () {
```
knaeckeKami, line 8: please remove the $ from the test strings.

I like static test names better if possible, dynamic ones make filtering by test name harder, and the test names generally become weird:

Screenshot 2024-02-20 at 17 51 10

knaeckeKami: Screenshot 2024-02-20 at 17 52 59

Also breaks IntellIj "retun test" feature

codegen/gql_tristate_value/lib/src/value.dart
```diff
@@ -14,6 +14,15 @@ sealed class Value<T extends Object> {
   /// If the value is non-null, it will be serialized as the value.
   const factory Value.present(T? value) = PresentValue<T>;
 
+  /// If the value is non-null, it will be serialized as the value.
+  /// If the value is null. It will not be serialized.
+  factory Value.ofNullable(T? value) {
+    if (value == null) {
+      return AbsentValue();
```
knaeckeKami, line 21: can this be const?

knaeckeKami:
final nit: 

Thinking about it, I think `.ofNullable` is not a great name for the behavior "present if not null, absent otherwise".

What about `Value.absentWhenNull()` ? of can you think of something more descriptive?

codegen/gql_tristate_value/lib/src/value.dart
LiLatee: It should be... My mistake. I am used to having a lint `prefer_const_constructors` 😅  Thanks 🙏

codegen/gql_tristate_value/test/value_test.dart
LiLatee: For sure `${Value.ofNullable}` should not be there. I forgot to check how it looks like in logs.
But personally I like to use e.g. `$AbsentValue`, because it's very helpful when refactoring the code. You don't have to remember about every place in the code where you have just a string with the class name.
But I changed all the names to static ones 👌

LiLatee:
Regarding the name I was thinking about `Value.valueOrAbsent`, but I have to admit that your `absentWhenNull` looks better 👌 Changed

codegen/gql_tristate_value/lib/src/value.dart
knaeckeKami: yeah I know, gql_pedantic should be modernized at some point ;)

knaeckeKami:
thanks! lgtm

PR #451, opened 2024-03-05
chore(vars-create-input): Adding Create Factories to Input types

caffeineflo:
Hello 👋 

it's the `create` factories guy again! 

Since we've been busy using the create factories we've added a while back, we noticed that we must have forgotten to apply the same factories to input types *doh* 
This makes it hard using the `create` factories/initializer very hard since input types are kinda necessary to be covered when your `Vars`/Inputs  go beyond simple value types.

This PR fixes that previous oversight and adds `create` factories to Input types.

I've added a unit test as well for completeness, but please let mw know if there are any questions on this! 
Thanks

### DIFF
--- a/codegen/end_to_end_test/build.yaml
+++ b/codegen/end_to_end_test/build.yaml
@@ -25,6 +25,7 @@ targets:
       gql_build|schema_builder:
         enabled: true
         options:
+          vars_create_factories: true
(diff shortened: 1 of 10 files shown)


knaeckeKami:
Hi!

Thanks, this feature looks good to me!

Issue #453, opened 2024-04-03
Printer fails with string blocks


sh1l0n:
When a String is on format

"Lore ipsum
One
Two"

The function **visitStringValueNode** should transform to

"Lore ipsum\nOne\nTwo"

knaeckeKami:
According to https://spec.graphql.org/June2018/#sec-String-Value
A multiline like ```
"Lore ipsum
One
Two"
```
is not allowed, for multiline strings, block strings (with triple quotes) are required.

Does this still fail for proper block strings?

**Comment by sh1l0n** [comment]
> According to https://spec.graphql.org/June2018/#sec-String-Value A multiline like
> 
> ```
> "Lore ipsum
> One
> Two"
> ```
> 
> is not allowed, for multiline strings, block strings (with triple quotes) are required.
> 
> Does this still fail for proper block strings?

I will show you with an example
1. The stringValueNode parameter contains this text: "Lore ipsum\nOne\nTwo";
2. stringValueNode.isBlock will return false since is not """
3. When printNode is called, stringValueNode will be displayed as
```string
"Lore ipsum
One
Two"
```
causing a problem when encoding the query with the gql_link RequestSerializer https://github.com/gql-dart/gql/blob/2dad214daebcc40e5b34a7315e1d4bfb85cb04d0/links/gql_link/lib/src/request_serializer.dart#L17

sh1l0n:
> According to https://spec.graphql.org/June2018/#sec-String-Value A multiline like
> 
> ```
> "Lore ipsum
> One
> Two"
> ```
> 
> is not allowed, for multiline strings, block strings (with triple quotes) are required.
> 
> Does this still fail for proper block strings?

Is is possible also that the Token is not computing the strings correctly? https://github.com/gql-dart/gql/blob/2dad214daebcc40e5b34a7315e1d4bfb85cb04d0/gql/lib/src/language/lexer.dart#L53

knaeckeKami:
> Is is possible also that the Token is not computing the strings correctly?

If you point me to any nontrivial piece of code and ask me "is it possible that there is a bug" I'll always say yes ;) 

Do you have any specific lead to a potential issue?

knaeckeKami:
> > According to https://spec.graphql.org/June2018/#sec-String-Value A multiline like
> > ```
> > "Lore ipsum
> > One
> > Two"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
[...]

PR #454, opened 2024-04-03
Fixed visitStringValueNode for string blocks

sh1l0n:
Fix: https://github.com/gql-dart/gql/issues/453


knaeckeKami:
Please fix the linting errors

Also, please add tests that call printNode(StringValueNode(...)) on blocks, non-blocks mit multiline strings.
I think the escape logic is now wrong, since you escape block strings which should not be escaped.

PR #462, opened 2024-06-05
Upgrade to web_socket_channel 3.0.0

catapop84:
This PR will upgrade to web_socket_channel 3.0.0


catapop84:
Hello @knaeckeKam, I saw. Trying to fix it. :)

catapop84:
@knaeckeKami , On my local pc test is valid. I'm using Dart SDK version: 3.4.1 (stable). That error might be from sdk. Can you check?

knaeckeKami:
It fails on the format check.
Please try executing `dart format` with 

Dart SDK version: 3.4.0 (stable)

catapop84:
@knaeckeKami I did format. it fails on auto reconnect. not sure why because on my local comp. is working. It's probably an OS problem. I'll test it on linux when I get home to further investigate the problem.

knaeckeKami:
FWIW it also fails locally for me, in MacOS

knaeckeKami:
possibly related: https://github.com/dart-lang/web_socket_channel/pull/370

PR #471, opened 2024-09-20
fix(gql_websocket_link): Fix race condition causing exception and double complete

james-pellow:
There is a race condition in the `subscribe` function of the _Client class.  If the dispose function returned by the `subscribe` method is called which calls the `releaser()` method, and then a `CompleteMessage` is received on the websocket while the `releaser()` is awaiting to compose its own `CompleteMessage`, then `releaser()` is called twice and the request is double released causing an exception.  I've added a flag to avoid a double release.

The second patch simply avoids the situation where a websocket is already closed when an error occurs, causing an already completed exception when `completeError()` is called on the completer.


knaeckeKami:
Hi! 
Thank you for this PR and sorry for the late response.

I just merged #466, which now causes a merge conflict. Can you take a look please?

PR #472, opened 2024-10-11
Wrap dio_link executeDioRequest in try-catch block

amritnew:
The request function has a return type of `Stream<Response>`. As a result, the catch block will use `Stream.error()` to asynchronously propagate the exception.

### DIFF
--- a/links/gql_dio_link/lib/src/dio_link.dart
+++ b/links/gql_dio_link/lib/src/dio_link.dart
@@ -67,7 +67,8 @@ class DioLink extends Link {
 
   @override
   Stream<Response> request(Request request, [forward]) async* {
-    final dio.Response<Map<String, dynamic>> dioResponse =
+    try {
+      final dio.Response<Map<String, dynamic>> dioResponse =
         await _executeDioRequest(
       request: request,
       headers: <String, String>{
@@ -96,6 +97,10 @@ class DioLink extends Link {
       response: gqlResponse.response,
       context: _updateResponseContext(gqlResponse, dioResponse),
     );
+    }
+    catch (e) {
+      yield* Stream.error(e);
+    }
   }
 
   dynamic _prepareRequestBody(Request request) {


knaeckeKami:
can you show in code how this would actually result in a different behavior? from what i see, this behaves exactly as before

anshul-arora:
@knaeckeKami 
The above fix seems to fix the issue i was facing with bad network connectivity

What i see from the code is that `_executeDioRequest` method was throwing an exception, which was not being catched and yielded by the request method

thanks @amritnew for the fix 

` Fatal Exception: FlutterError
LinkException(DioException [connection error]: The connection errored: Failed host lookup: 'xxx.xxx.xx.com' This indicates an error which most likely cannot be solved by the library. Error: SocketException: Failed host lookup: 'xxx.xxx.xx.com' (OS Error: nodename nor servname provided, or not known, errno = 8), #0 DioMixin.fetch (package:dio/src/dio_mixin.dart:520) <asynchronous suspension> #1 DioLink._executeDioRequest (package:gql_dio_link/src/dio_link.dart:198) <asynchronous suspension> #2 DioLink.request (package:gql_dio_link/src/dio_link.dart:71) <asynchronous suspension> ) `

AngeloAvv:
Hey @knaeckeKami, do you think you'll be able to merge this soon?

knaeckeKami:
I honestly cannot see any difference in behaviour between the old state and the new state (wrapping in try/catch) - 
Dart already converts exceptions to stream errors in `async*` functions!

The only difference is that one would lose the Stack trace when doing that, since you do not pass the stacktrace from the exception to `Stream.error`.

This could be fixed by doing

```dart
  catch (e, stackTrace) {
    yield* Stream.error(e, stackTrace);
  }
```

But this would just replicate the exact behaviour that Dart would already provide out of the box

Do I misunderstand something?

AngeloAvv:
> The only difference is that one would lose the Stack trace when doing that, since you do not pass the stacktrace from the exception to `Stream.error`.

@knaeckeKami yes, you're right, but for some reason, the error is being raised because if you don't declare the onError when listening to the stream, it will be handled by the topmost handler.

In my case, I'm getting this error on crashlytics when the server is temporary unavailable:

``` 
          Fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: DioLinkServerException(originalException: DioException [bad response]: This exception was thrown because the response has a status code of 504 and RequestOptions.validateStatus was configured to throw for this status code.
The status code of 504 has the following meaning: "Server error - the server failed to fulfil an apparently valid request"
Read more about status codes at https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
In order to resolve this exception you typically have either to verify and fix your request code or you have to fix the server code.
, status: 504, response: ...
       at DioLink._executeDioRequest(dio_link.dart:246)
       at DioLink.request(dio_link.dart:71)

```

If you want, we can discuss [here](https://github.com/gql-dart/gql/issues/473)

knaeckeKami:
I think these stream errors would trigger the top level exception handler?


```dart
Stream a() {
  return Stream.error("oh no");
}

Stream b() async* {
  yield* Stream.error("oh no 2");
}

Stream c() async* {
  throw "oh no 3";
}

void main() {
  //a().listen((_) {}); // returning stream.error
  //b().listen((_) {});  // yielding Stream.error, swallows stracktrace but still calls exception handler
 c().listen((_) {}); // current behaviour
}

```

AngeloAvv:
> I think these stream errors would trigger the top level exception handler?

Exactly, but since I'm not using .listen directly, there should be a missing error handling somewhere.
My stack relies on graphql, graphql_codegen and gql_dio_link, basically I'm using these three libraries to interact with my gql endpoint with both http and ws protocols. Http for queries and mutations, and ws for subscriptions. To redirect the request according to the type, I'm using Link.split as follows:

``` dart
Provider<GraphQLClient>(
    create: (context) {
      final wssLink = context.read<WssCustomLink>();

      return GraphQLClient(
        /// **NOTE** The default store is the InMemoryStore, which does NOT persist to disk
        cache: GraphQLCache(),
        link: Link.split(
          (request) => request.isSubscription,
          wssLink,
          DioLink(K.graphqlUrl, client: context.read<Dio>()),
        ),
      );
    },
  ),
```

I created a custom Link (WssCustomLink) to handle token refresh, as suggested [here](https://github.com/zino-hofmann/graphql-flutter/tree/main/packages/graphql#updating-websocket-socket-connection)

Do you think it's something on your side, or should I open a ticket [here](https://github.com/zino-app/graphql-flutter/)?

I'm sorry I necroposted here, but the stacktrace pointed me right into this PR and the attached issue

knaeckeKami:
I don't really know where the missing error handling in your app or `graphql` or `graphql_flutter` might be, but merging this PR would not change anything there as far as I can see, except for getting rid of the stack trace. 

I suggest opening an issue in graphql-flutter.

PR #481, opened 2025-03-04
Introducing melos to replace multipack in the monorepo

caffeineflo:
Hi 👋 ,

To make contributions to this repo easier, I thought I'll take the time and introduce `melos` to this repo.

For those that don't know what `melos` is, it's a monorepo manager suited to help managing different packages/dependencies as well as aiding with local development. More info can be found here https://melos.invertase.dev

This PR adds the `melos.yaml` to the root to configure melos and adds melos as a dev dependency to the root as well. I've added some info to the `README.md` as well, to help new contributors.

I think this change makes sense, given that `multipack` hasn't been updated or maintained in some time and `melos` has become the standard for dart monorepos! 

Let me know what y'all think. 

Update:

I have replace multipack on CI as well and made some changes that I think make sense. They are well documented on the README.md now. In essence:

* To have a commit history with version changes and enabling to "graduate" an alpha release to a patch/minor/major release, a maintainer now needs to push or merge pubspec changes to either the `release` or `main` branch.
* Commands to do that conveniently are available via `melos run bump-alpha/patch/minor/major`. The workflows also work if pubspec changes are made manually
* The dry-run workflow still does a dry-run but doesn't ignore errors anymore 

I don't suspect for the release to run into any issues, but wasn't able to test the same due to missing permissions to push to pub dev. Happy to take a look on the next release if anything should fail.


knaeckeKami:
Hi! 

Thanks for this PR!

I think this a great change.

The global pubspec.yaml file breaks multipack on the CI. I think we should either do the migration all at once or work around that be deleting the global pubspec.yaml file in the CI so multipack does not crash.

caffeineflo:
> Hi!
> 
> Thanks for this PR!
> 
> I think this a great change.
> 
> The global pubspec.yaml file breaks multipack on the CI. I think we should either do the migration all at once or work around that be deleting the global pubspec.yaml file in the CI so multipack does not crash.

You are faster than I can make the changes 😃 
I am working on making the changes all at once, will have an update shortly!

caffeineflo:
@knaeckeKami This should be good now. I added some more info to the description and the README.md

I tried to capture all the packages and make them known in the `melos.yaml` - hope I didn't forget one. Also had to add a small fix for contributions coming from forks, which works now

knaeckeKami:
LGTM!

PR #485, opened 2025-03-31
Stacked PR 2: GQL Code Builder Refactor to enable Interface/Fragment Type Hierarchy

caffeineflo:
This PR Stacks on top of #484  - with #484 being the base for this PR. Individually review each PR and then merge this into #484 before merging #484.

Explanation of changes:
# Refactoring the GraphQL Code Generator: Interface Fragment Support

## The Problem Solved in First PR

Our first PR addressed a critical issue with GraphQL fragments on interface types. When a fragment was defined on an interface with inline fragments for specific implementations (like `BookFragment` on the `Book` interface with specific fragments for `Textbook` and `ColoringBook`), the generated Dart code would fail at runtime due to missing implementation classes and improper type hierarchies.

This resulted in runtime cast exceptions in the `when` extension method because nothing actually implemented interfaces like `GBookFragment__asTextbook`. Additionally, type-specific data would be lost during deserialization.

## Why This Refactoring PR Was Necessary

While the first PR fixed the immediate issues, it revealed deeper structural problems in the codebase:

1. **File Size and Complexity**: The original `data.dart` had grown to over 700 lines with complex, deeply nested logic making it increasingly difficult to debug and maintain.

2. **Tightly Coupled Logic**: Different concerns were mixed together in the same functions, making it hard to reason about the code or modify specific parts without unintended side effects.

3. **Insufficient Fragment Deduplication**: The initial fix still had issues with duplicate fragment generation in nested queries, evidenced by the `nested_duplicate_fragments.graphql` test case.

4. **Limited Documentation**: The codebase lacked clear explanations of the generation process, making it challenging for new contributors to understand the logic.

## How We Implemented the Refactoring

This stacked PR takes a systematic approach to improving the codebase:

1. **Modular Architecture**: We extracted related functionality into dedicated, focused files:
   - `class_utils.dart`: Responsible for building the main class structures
   - `field_utils.dart`: Handles field processing and getter generation
   - `fragment_utils.dart`: Manages fragment expansion and type resolution
   - `selection_utils.dart`: Focuses on selection merging and expansion logic

2. **Enhanced Fragment Deduplication**: We implemented a fragment source tracking system in `buildNestedFieldClasses()` that detects when multiple field classes come from the same fragment source, preventing duplicate class generation and using aliases instead.

3. **Improved Type Mapping**: Created a more comprehensive type mapping system that consistently tracks the relationship between GraphQL types and generated Dart classes.

4. **Better Documentation**: Added extensive documentation to explain the purpose and behavior of each component and function, making it easier for future contributors to understand the codebase.

5. **Granular Testing**: Added the `nested_duplicate_fragments.graphql` test to verify our fix works for complex nested fragment scenarios.

## Results and Benefits

This refactoring provides several important benefits:

1. **Fixed Fragment Duplication**: The `nested_duplicate_fragments.graphql` test case now passes, showing that we correctly handle fragment deduplication.

2. **Improved Maintainability**: Breaking down the monolithic file makes it easier to find, understand, and modify specific parts of the code generation process.

3. **Better Debugging**: With clearer separation of concerns, issues are easier to isolate and fix.

4. **More Robust Implementation**: The refactored code handles more edge cases correctly and is more resilient to complex GraphQL schemas.

5. **Enhanced Extensibility**: The modular structure makes it easier to add new features or optimization in the future.

These changes represent a substantial improvement to the codebase that not only fixes the immediate issue with interface fragments but also lays a stronger foundation for future development.


codegen/end_to_end_test/lib/graphql/__generated__/serializers.gql.g.dart
```diff
@@ -167,6 +291,16 @@ Serializers _$serializers = (new Serializers().toBuilder()
...
+      ..addBuilderFactory(
+          const FullType(BuiltList, const [
+            const FullType.nullable(GheroFieldsFragment__asHuman_friends)
+          ]),
+          () => new ListBuilder<GheroFieldsFragment__asHuman_friends?>())
+      ..addBuilderFactory(
+          const FullType(BuiltList, const [
+            const FullType.nullable(GheroFieldsFragment__asHuman_friends)
+          ]),
+          () => new ListBuilder<GheroFieldsFragment__asHuman_friends?>())
```
caffeineflo, line 303: These entries shouldn't happen and this is the one big remaining issue in this PR.

These factories get added due `GheroFieldsFragment__asHuman_friends` having a list of `friends` that are of type `BuiltList` which `built_value` references as needing to be added as a `BuilderFactory`

```
abstract class GheroFieldsFragment__asHuman
    implements GheroFieldsFragment, GhumanFieldsFragment {
  @override
  String get G__typename;
  @override
  String get id;
  @override
  String get name;
  @override
  String? get homePlanet;
  @override
  BuiltList<GheroFieldsFragment__asHuman_friends?>? get friends;
}
```

The log output of `gql_build` even says that we are missing an import and that one will be added, but this is actually untrue and none will be added
```
[INFO] gql_build:serializer_builder on lib/$lib$:Forcing import reference for GheroFieldsFragment__asHuman_friends from package:end_to_end_test/fragments/__generated__/hero_with_interface_subtyped_fragments.data
```

I haven't actually found out why that is the case (that the import does not get added), but happy to hear if someone else might have an idea why.

In either case, I wasn't able to find a way to fix the issue yet. Afaik, the type hierarchy that the codegenerator now generates is correct and this factory should not be added since all the classes that implement the abstract `GheroFieldsFragment__asHuman_friends` class actually override the `friends` list with a builder type. I couldn't find a way to make `built_value` behave here. 

I can see two options to go and solve this now:

1. Fix the import logic, so this class will get imported (and shown). While we wouldn't need this entry in first place, I can't forsee this causing a lot of issues.
or
2. Have a cleanup phase that removed abstract classes that are not instantiable like this one from the generated `serializers.gql.g.dart` file

caffeineflo: I actually implemented 1 on my own fork by just importing the whole file. So that is another option

knaeckeKami:
Thanks! 
This look great.

It's a big PR, and it might take me a few days to review this.

codegen/end_to_end_test/lib/graphql/__generated__/serializers.gql.g.dart
knaeckeKami: Hm, I can't reproduce the `Forcing import reference` locally. Can you give me a instruction on how to generate this warning?

knaeckeKami: Ah, I see you added another commit for importing the whole file. I think that's fine since there should not be any duplicate names

knaeckeKami:
Thank you very much for this! 

I think this a great improvement.

There are some things that could be improved in the future - like the assumptions on the name of the generated classes which by now are scattered over the code base - but I think for now this is good.

PR #491, opened 2025-05-21
feat(gql_socket_link): BREAKING CHANGE multiplex websocket connection

agufagit:
@knaeckeKami I'm making a breaking change in graphql_transport_ws.dart, please take a look

### DIFF
--- a/links/gql_websocket_link/CHANGELOG.md
+++ b/links/gql_websocket_link/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.0.0
+
+- BREAKING: graphql-transport-ws: streaming operations and single result operations are now multiplexed on a single connection, user has to manually resubscribe subscription when connection is broken
+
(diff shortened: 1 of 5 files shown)


knaeckeKami:
Can you explain the change you did a bit more? From what I understand, there has always been just one WebSocket connection, no? And the code changes mostly add commented out code, I don't see any actual change in behaviour?

agufagit:
It's actually the result of a series of pull requests that I made.  Auto resubscribe may or may not work when app is in foreground because of strange behaviors of web_socket_channel 3.0.3.  Auto resubscribe definitely not work when app is in background (lock screen).  So for this version, user has to manually resubscribe, I think it would be best to notify users that they need to manually resubscribe.  And single-result-operation shouldn't be in for loop, it should complete on NextMessage instead of CompleteMessage, otherwise it might get re-sent.

PR #497, opened 2025-09-01
Fix: JsonObject fields should not use replace() method with tristate_optionals

### LINKED ISSUE #649: Error: The method 'replace' isn't defined for type 'JsonObject?' when tristate_optionals is enabled
### Description
When using Ferry with `JsonObject` type (from `built_value` package) and **`tristate_optionals: true`** option enabled, the generated serializer code contains calls to `.replace()` method on `JsonObject?` fields, but `JsonObject` doesn't have a `replace` method.

**Important: This error only occurs when `tristate_optionals: true` is set in the build configuration.**

### Environment
- **Ferry version**: Latest
- **built_value version**: 8.11.1
- **Flutter/Dart SDK**: Latest stable

### Reproduction Steps

1. Define a GraphQL schema with JSON scalar type:
```graphql
scalar JSON
```

2. Configure `build.yaml` with type override **and tristate_optionals enabled**:
```yaml
targets:
  $default:
    builders:
      ferry_generator|graphql_builder:
        options:
          schema: path/to/schema.graphql
          tristate_optionals: true  # ← This triggers the issue
          type_overrides:
            JSON:
              name: JsonObject
              import: "package:built_value/json_object.dart"
```

3. Generate code using `build_runner`

### Generated Code Issue

The generated serializer code produces:
```dart
// In schema.schema.gql.dart
case 'params':
  var _$fieldValue = serializers.deserialize(value,
      specifiedType: const FullType(_i1.JsonObject)) as _i1.JsonObject;
  builder.params.replace(_$fieldValue);  // ❌ Error here
  break;

case 'body':
  var _$fieldValue = serializers.deserialize(value,
      specifiedType: const FullType(_i1.JsonObject)) as _i1.JsonObject;
  builder.body.replace(_$fieldValue);  // ❌ Error here
  break;
```

### Error Message
```
Error: The method 'replace' isn't defined for type 'JsonObject?'.
 - 'JsonObject' is from 'package:built_value/json_object.dart'
Try correcting the name to the name of an existing method, or defining a method named 'replace'.
              builder.params.replace(_$fieldValue);
                             ^^^^^^^
```

### Expected Behavior
The generated code should use proper assignment for `JsonObject` fields instead of calling `.replace()` method, similar to how other non-built types are handled:

```dart
// Should be:
builder.params = _$fieldValue;
// Instead of:
builder.params.replace(_$fieldValue);
```

### Additional Context
- The issue **does not occur** when `tristate_optionals: false` (or omitted)
- Looking at the generated `.g.dart` file, the builder has simple setters for `JsonObject` fields:
```dart
_i1.JsonObject? _params;
_i1.JsonObject? get params => _$this._params;
set params(_i1.JsonObject? params) => _$this._params = params;
```

The issue seems to be that when `tristate_optionals` is enabled, the code generator incorrectly treats `JsonObject` as a built collection type that has a `replace` method, when it should treat it as a simple value type.

### Workaround
Disabling `tristate_optionals` avoids the issue, but this is not ideal when you need to distinguish between null and absent values in GraphQL inputs.

devunt:
## Description
This PR fixes an issue where `JsonObject` fields incorrectly generate calls to `.replace()` method when `tristate_optionals` is enabled.

## Problem
When using Ferry with `JsonObject` type (from `built_value` package) and `tristate_optionals: true` option enabled, the generated serializer code contains calls to `.replace()` method on `JsonObject?` fields. However, `JsonObject` doesn't have a `replace` method, causing compilation errors.

## Solution
Added special handling for `JsonObject` type in the tristate optionals code generator to ensure it's treated as a simple value type (using direct assignment) rather than a built collection type (which would use `.replace()` method).

## Changes
- Modified `/codegen/gql_code_builder/lib/src/tristate_optionals.dart` to detect `JsonObject` type and exclude it from being treated as a builder type

## Testing
Tested with a real-world Flutter project using:
- `tristate_optionals: true`
- `JsonObject` type override for JSON scalar
- Verified that generated code now compiles successfully

## Related Issues
- Fixes gql-dart/ferry#649

## Before/After

**Before (causes compilation error):**
```dart
case 'params':
  var _$fieldValue = serializers.deserialize(value,
      specifiedType: const FullType(_i1.JsonObject)) as _i1.JsonObject;
  builder.params.replace(_$fieldValue);  // ❌ Error: replace() not defined
  break;
```

**After (works correctly):**
```dart
case 'params':
  var _$fieldValue = serializers.deserialize(value,
      specifiedType: const FullType(_i1.JsonObject)) as _i1.JsonObject;
  builder.params = _$fieldValue;  // ✅ Direct assignment
  break;
```

### DIFF
--- a/codegen/gql_code_builder/lib/src/tristate_optionals.dart
+++ b/codegen/gql_code_builder/lib/src/tristate_optionals.dart
@@ -187,8 +187,14 @@ String _generateFieldDeserializers(
           getTypeDefinitionNode(schemaSource.document, originalSymbolName);
 
       //TODO this feels flaky, find a better way
+      // Check if type is a built collection (has replace method)
+      // JsonObject from built_value doesn't have replace method
+      final isJsonObject = type.symbol == "JsonObject" && 
+          type.url == "package:built_value/json_object.dart";
+      
       final isBuilder = type.url != null &&
           !isWrappedValue &&
+          !isJsonObject &&
           (typeDefNode is! ScalarTypeDefinitionNode &&
               typeDefNode is! EnumTypeDefinitionNode);
 


knaeckeKami:
Thanks, looks good to me as workaround for now . Need to look into a proper fix later.

knaeckeKami:
Please fix the formatting though

PR #501, opened 2025-10-30
Fixes issue with multiple nested fragments on interface and union

### LINKED ISSUE #610: Fragments not working with interfaces
When I create a Fragment with a type which is an interface, or has fields which are interfaces, the generated code does not work at runtime.

For example let's have this schema:
```graphql
interface Book {
  title: String!
  author: String!
}

type Textbook implements Book {
  title: String!
  author: String!
  courses: [String!]!
}

type ColoringBook implements Book {
  title: String!
  author: String!
  colors: [String!]!
}

type Query {
  books: [Book!]!
}
```

And this query:
```dart
fragment BookFragment on Book {
  title
  ... on Textbook {
      courses
  }
  ... on ColoringBook {
      colors
  }
}

query GetBooks {
  books {
    ...BookFragment
  }
}
```

These parameters are set on build.yaml:
```yaml
  when_extensions:
    when: true
    maybeWhen: true
```

When we generate the code, we can see that nothing implements `GBookFragment__asTextbook`, yet we have this kind of generted code:
```dart
extension GBookFragmentWhenExtension on GBookFragment {
  _T when<_T>({
    required _T Function(GBookFragment__asTextbook) textbook,
    required _T Function(GBookFragment__asColoringBook) coloringBook,
    required _T Function() orElse,
  }) {
    switch (G__typename) {
      case 'Textbook':
        return textbook((this as GBookFragment__asTextbook));
      case 'ColoringBook':
        return coloringBook((this as GBookFragment__asColoringBook));
      default:
        return orElse();
    }
  }
}
```

Since nothing implements `GBookFragment__asTextbook`, the cast `(this as GBookFragment__asTextbook)` will fail.
Is there a workaround for that?

letsar:
This PR aims to fix https://github.com/gql-dart/ferry/issues/610.
I don't fully understand the code, so I'm not sure why I needed to use `superclassSelections` instead of `context.hierarchy` in `inline_fragment_classes.dart`, but it seems to work. I've tested it in our codebase and so far I don't see any regression and we can now use the kind of fragments discussed in the issue 🥳 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ doc/api/
 
 **/test/**/*.mocks.dart
 pubspec_overrides.yaml
+
+examples/gql_example_flutter/ios/Flutter/ephemeral/
--- a/codegen/end_to_end_test/lib/fragments/__generated__/hero_with_interface_subtyped_fragments.data.gql.dart
+++ b/codegen/end_to_end_test/lib/fragments/__generated__/hero_with_interface_subtyped_fragments.data.gql.dart
@@ -343,8 +343,7 @@ abstract class GHeroWithInterfaceSubTypedFragmentsData_hero__asHuman_friends__as
             GHeroWithInterfaceSubTypedFragmentsData_hero__asHuman_friends__asDroidBuilder>,
         GheroFieldsFragment__asHuman_friends,
         GhumanFieldsFragment_friends,
-        GheroFieldsFragment,
-        GheroFieldsFragment__asDroid,
+        GheroFieldsFragment__asHuman_friends__asDroid,
         GhumanFieldsFragment_friends__asDroid,
         GHeroWithInterfaceSubTypedFragmentsData_hero__asHuman_friends,
         GdroidFieldsFragment {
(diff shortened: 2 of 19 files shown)


knaeckeKami:
Thanks so much for this!

I am currently trying to understand _why_ this would work ;)

Oh...

```dart
void addInterface(String name) {
    final parts = name.split("__");           // Split on "__"
    final base = parts[0];                    // e.g., "GHeroData_hero"
    hierarchy.putIfAbsent(base, () => {});    // Create entry for base
    if (parts.length > 1) {                   // If specialized version
      hierarchy[base]!.add(name);             // Add to base's set
    }
  }
```

I think the culprit is just the naive split on the first `__`, which breaks down in nested scenarios ...

letsar:
Hello, I don't have the time right now to explain a little further, but this fix introduced a regression when we use a named Fragment on a specific union type, which select a field which is an interface. I didn't have a simple repro case for now, but I wanted to warn you about this. I didn't look at the `addInterface` method, but yes maybe, the solution is there after all.

knaeckeKami:
Thanks! I already stumbled upon this or something similar when trying to write more comprehensive regression tests for this - which is why I did not merge this yet. did not have time to look into it further yet

